PR #201, opened 2021-11-14
Better CI

MOZGIII:
This is mainly about trying to smartly detect feature flags to test with and testing independent crates in isolation with all meaningful feature flag combinations. This work is not complete, stuck mainly at the lack of metadata to determine what would be the meaningful combinations.


MOZGIII:
This is obsolete, the CI have moved way forward past this; what's still lacking is proper atomic crate/feature tests and dedicated wasm tests - but that work has to be redone from scratch either way.

PR #408, opened 2022-06-27
EVM accounts lookup

MOZGIII:
This PR updates the EVM account mapping logic to take the EVM account mappings into account.

---

To do:

- [x] `frame_system::Config::Lookup` is made to use the mapping
- [x] `pallet_evm::Config::CallOrigin` and `pallet_evm::Config::WithdrawOrigin` is made to use the system lookup
- ~use a fallible EVM to make the address lookups fallible in the EVM internals~ postpone it to a separate PR
- [x] ~set the `pallet_evm::Config::AddressMapping` to the fallible lookup logic via the mapping~ we won't have a fallible EVM at this PR, so postpone this too; instead, we have implemented a fallback to the wrapped eth 20-byte address in the "native" `AccountId32`; see #659
- [x] tweak `pallet_evm::Config::FindAuthor` accordingly
- [x] tweak `pallet_evm::Config::OnChargeTransaction` to handle the tips same way we handle the tips
- ~~extract the implementation into a separate crate (or not?)~~ we will leave this as is for now, and refactor the runtime later when we dedicate the time specifically for it
- [ ] implement the tests with a mock runtime

### DIFF
--- a/crates/humanode-runtime/src/evm_address.rs
+++ b/crates/humanode-runtime/src/evm_address.rs
@@ -174,3 +174,29 @@ where
         Value::get()
     }
 }
+
+pub struct FindAuthorLookup<T, F>(core::marker::PhantomData<(F, T)>);
+
+impl<T, F> FindAuthor<H160> for FindAuthorLookup<T, F>
+where
+    T: frame_system::Config,
+    <T as frame_system::Config>::Lookup: StaticLookup<Source = MultiAddress<T::AccountId, ()>>,
+    T: pallet_evm_accounts_mapping::Config,
+    F: FindAuthor<T::AccountId>,
+{
+    fn find_author<'a, I>(digests: I) -> Option<H160>
+    where
+        I: 'a + IntoIterator<Item = (ConsensusEngineId, &'a [u8])>,
+    {
+        let account_id = F::find_author(digests)?;
+
+        match T::Lookup::unlookup(account_id) {
+            MultiAddress::Address20(eth_address) => Some(H160(eth_address)),
+            MultiAddress::Id(account_id) => {
+                pallet_evm_accounts_mapping::Pallet::<T>::ethereum_addresses(account_id)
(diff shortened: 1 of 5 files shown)


crates/humanode-runtime/src/evm_address.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    fn lookup(s: Self::Source) -> Result<Self::Target, frame_support::error::LookupError> {
+        match s {
+            // Pass "native" address directly as-is.
+            MultiAddress::Id(id) => Ok(id),
+            // Map 20-byte address to a native address via ethereum address mapping pallet,
+            // if the mapping exists.
+            // If the mapping does not exist, we still want to proceed without an error and map the
+            // address to something. Well, the way we do this now is invoking
+            // the [`pallet_evm::HashedAddressMapping::<BlakeTwo256>`].
```
dmitrylavrenov, line 33: Would like to get understanding of allowing this. What about the future possible mapping of this `Address20` with another `AccountId`?

crates/humanode-runtime/src/evm_address.rs
```diff
@@ -0,0 +1,202 @@
...
+        // panic to kill the whole EVM invocation.
+        // Ideally we'd just return an error here instead, but the signature of this trait fn
+        // does not allow it.
+        panic!(
+            "lookup failed for evm address {address}; this is not a bug, you are just accessing the unmapped evm address",
+        );
+    }
+}
+
+/// A [`pallet_evm::AddressMapping`] implementation that logs a warining and always returns
```
dmitrylavrenov, line 156:
```suggestion
/// A [`pallet_evm::AddressMapping`] implementation that logs a warning and always returns
```

dmitrylavrenov (approved):
Understandable with good explanation details. Would like to discuss some things.

MOZGIII:
This is obsolete with our new EVM integration model

PR #614, opened 2023-03-27
Weight to fee multiplier test improvements

MOZGIII:
This PR implements all the necessary utilities to replace the hardcoded length with a properly computed one from a real extrinsic constructed in the test.


dmitrylavrenov (approved):
Awesome improvements :)

PR #616, opened 2023-03-28
E2E test for weight to fee multiplier

### LINKED ISSUE #613: E2E test for weight to fee multiplier
A follow-up after #585.

We need to run the test at CI.

dmitrylavrenov:
Closes #613


MOZGIII (changes requested):
We should not hardcode the tx binary, but rather construct one on the fly. The rationale is the actual tx might change over time (if we add more signed extras for instance). Then the hardcoded tx will no longer be relevant, but we want it to take care of itself automatically.

Overall, it seems like we'd need to do it via Polkadot.js, in JavaScript (TypeScript). This, in turn, would require adding a whole new test suite and CI flow to support TypeScript tests.

dmitrylavrenov:
Discussed and decided to add a whole new test suite and CI flow to support TypeScript tests. So, this code is not relevant.

PR #633, opened 2023-04-04
Try-runtime e2e test

dmitrylavrenov:
- [x] e2e job using matrix strategy to qualify `humanode-peer` build args and tests to be run
- [x] `execute-block` test
- [x] `on-runtime-upgrade` test


utils/e2e-tests/tests/try-runtime-check.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Make temporary test directory.
+TEMPDIR="$(mktemp -d)"
+trap 'rm -rf "$TEMPDIR"; pkill -P "$$"' EXIT
+
+# Run the node.
+"$COMMAND" --dev --base-path "$TEMPDIR" &
+
+# Kepp the node running to have around 3 finalized blocks.
+sleep 40
```
MOZGIII, line 15: Do not use sleep as synchronization method in tests. Instead, run a query in an infinite loop with a smaller sleep (and exit when the whatever conditions are met, in the case 3 blocks are finalized), and apply an overall timeout for the operation.

MOZGIII (changes requested):
Very nice! There are some issues though, I left comments.

MOZGIII (changes requested):
Very nice! There are some issues though, I left comments.

MOZGIII (changes requested):
Very nice! There are some issues though, I left comments.

utils/e2e-tests/tests/try-runtime-check.sh
dmitrylavrenov: Improved

.github/workflows/code.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         timeout-minutes: 10
 
       - name: Build binaries
-        run: cargo build --release
+        run: cargo build --release --features try-runtime
```
MOZGIII, line 127: Oh, this is actually a problem. We need to run E2E tests for the regular builds...

MOZGIII:
We have similar issues here to what we had while we were initially working on the E2E tests with @tonycdot.

Issues so far:
- the child processes spawned during the test are not terminated reliably
- the timeout is implemented as in the cooperative scheduling, but we actually need to run it with preemptive scheduling in mind

MOZGIII (changes requested):
Ok, two things left here to solve this:
1. Qualify whether a test is a try-runtime or a regular one, and enable running the tests with only the selected qualifier.
2. Add a second job E2E to CI (instead of replacing the one we have currently) with the `try-runtime` build and E2E tests invocation. Converting the existing job into a matrix would actually be preferable.

This can really be finished quite easily and quickly.

.github/workflows/code.yml
dmitrylavrenov: Redesigned

.github/workflows/code.yml
```diff
@@ -72,7 +72,16 @@ jobs:
...
   e2e:
-    name: End-to-end tests
+    strategy:
+      matrix:
+        qualifier: ["base", "try-runtime"]
+        include:
+          - qualifier: "base"
+            extend-build-args: ""
+          - qualifier: "try-runtime"
+            extend-build-args: "--features try-runtime"
```
MOZGIII, line 82:
```suggestion
      matrix:
        test:
          - qualifier: "base"
            extend-build-args: ""
          - qualifier: "try-runtime"
            extend-build-args: "--features try-runtime"
```

Better use the matrix like this, it is easier to understand.

MOZGIII:
```suggestion
      matrix:
        test:
          - suite: "base"
            build-args: ""
          - suite: "try-runtime"
            build-args: "--features try-runtime"
```

And this is my further suggestion to tweak the key names

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/633/commits/d08d04f56d7852a9e1708985586cdb187f385e0b

.github/workflows/code.yml
```diff
@@ -91,13 +99,13 @@ jobs:
         timeout-minutes: 10
 
       - name: Build binaries
-        run: cargo build --release --locked
+        run: cargo build --release --locked ${{ matrix.test.build-args }}
 
       - name: Run bash e2e tests
-        run: utils/e2e-tests/bash/run-tests "target/release/humanode-peer"
+        run: utils/e2e-tests/bash/run-tests ${{ matrix.test.suite }} "target/release/humanode-peer"
```
MOZGIII, line 105:
```suggestion
        run: utils/e2e-tests/bash/run-tests "$TEST_SUITE" "target/release/humanode-peer"
        env:
           TEST_SUITE: ${{ matrix.test.suite }}
```

This is a bit safer if we consider quoting and shell injection

dmitrylavrenov: Got it, thanks, improved

MOZGIII: You got it slightly wong though - see the quoting around the env var

.github/workflows/code.yml
```diff
@@ -91,13 +99,19 @@ jobs:
...
 
       - name: Build binaries
-        run: cargo build --release --locked
+        run: cargo build --release --locked $BUILD_ARGS
+        env:
+          BUILD_ARGS: ${{ matrix.test.build-args }}
 
       - name: Run bash e2e tests
-        run: utils/e2e-tests/bash/run-tests "target/release/humanode-peer"
+        run: utils/e2e-tests/bash/run-tests $TEST_SUITE "target/release/humanode-peer" "target/release/wbuild/humanode-runtime/humanode_runtime.wasm"
```
MOZGIII, line 107:
```suggestion
        run: utils/e2e-tests/bash/run-tests "$TEST_SUITE" "target/release/humanode-peer" "target/release/wbuild/humanode-runtime/humanode_runtime.wasm"
```

utils/e2e-tests/bash/tests/try-runtime/on_runtime_upgrade_live.sh
MOZGIII, line 1: Please rename this file to match the naming convention of the rest

utils/e2e-tests/bash/tests/try-runtime/on_runtime_upgrade_live.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+HUMANODE_PEER_PATH="${1?Provide the path to the humanode peer as the first argument}"
+shift
+
+RUNTIME_WASM="${1?Provide the path to runtime wasm file as the second argument}"
```
MOZGIII, line 7: This might be better to pass via env vars...

MOZGIII (approved):
Great stuff! The overall approach has no issues - well, besides not reusing the builds and repeating the work that build jobs do - but is something to address in a separate PR.

I have left a couple of comments, all of them are small nits though.

utils/e2e-tests/bash/tests/try-runtime/on_runtime_upgrade_live.sh
MOZGIII: We can address this in a separate PR, as this would cause a refactor that's unrelated to this PR.

utils/e2e-tests/bash/tests/try-runtime/on_runtime_upgrade_live.sh
dmitrylavrenov: Can you suggest smth better ?)
My thoughts are that we have `try-runtime` test suite where we would like to test:
- `execute-block` subcommand
- `on_runtime_upgrade` subcommand with live networks.

That's why we have `execute-block.sh` and `on_runtime_upgrade_live.sh`

.github/workflows/code.yml
dmitrylavrenov: Got it, fixed - https://github.com/humanode-network/humanode/pull/633/commits/089f56bc417410c7d0ea60cd5c4964f83d9019db )

utils/e2e-tests/bash/tests/try-runtime/on_runtime_upgrade_live.sh
MOZGIII: Why `on_runtime_upgrade`? The subcommand is `on-runtime-upgrade`.

dmitrylavrenov: Ah, I got ))

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/633/commits/8d0c00dfe137b07c68cae72cd299fb1b2f42df3e

PR #641, opened 2023-04-09
Switch to a new rand and ed25519-dalek implementations

MOZGIII:
This is relevant to the robonode only; thus, it is a low priority since we will be working on the new implementation for it soon.

### DIFF
--- a/crates/humanode-runtime/src/robonode.rs
+++ b/crates/humanode-runtime/src/robonode.rs
@@ -11,7 +11,12 @@ impl PublicKey {
     pub fn from_bytes(
         bytes: &[u8],
     ) -> Result<Self, robonode_crypto::ed25519_dalek::ed25519::Error> {
-        let actual_key = robonode_crypto::PublicKey::from_bytes(bytes)?;
+        if bytes.len() != 32 {
+            return Err(robonode_crypto::ed25519_dalek::ed25519::Error::new());
+        }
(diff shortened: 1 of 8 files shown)


crates/humanode-runtime/src/robonode.rs
```diff
@@ -11,7 +11,12 @@ impl PublicKey {
     pub fn from_bytes(
         bytes: &[u8],
     ) -> Result<Self, robonode_crypto::ed25519_dalek::ed25519::Error> {
-        let actual_key = robonode_crypto::PublicKey::from_bytes(bytes)?;
+        if bytes.len() != 32 {
+            return Err(robonode_crypto::ed25519_dalek::ed25519::Error::new());
```
dmitrylavrenov, line 15: Is it possible to add more info into error?

dmitrylavrenov (approved):

crates/humanode-runtime/src/robonode.rs
MOZGIII: Unfortunately, this is not possible with their API. I think it is intentional. We could add our own wrapper error, but what would be want to put there?

dmitrylavrenov: I was asking just to get if it's possible to note that the reason is in bytes len.

PR #649, opened 2023-04-11
More bioauth authentication related events

dmitrylavrenov:
During `ImOnline` research and investigating authentications history I've come up that it would be great to have more events that will explicitly point on the reason the authentication is removed from the state.


crates/pallet-bioauth/src/lib.rs
```diff
@@ -400,6 +415,13 @@ pub mod pallet {
                     }
                     true
                 });
+                if removed {
+                    // Emit an event.
+                    Self::deposit_event(Event::AuthenticationRemoved {
+                        validator_public_key: public_key.clone(),
+                        reason,
+                    });
+                }
```
MOZGIII, line 424: This most likely should not be in the `ActiveAuthentications::mutate` block

crates/pallet-bioauth/src/lib.rs
```diff
@@ -331,6 +334,15 @@ pub mod pallet {
         NewAuthentication {
             validator_public_key: T::ValidatorPublicKey,
         },
+        /// The authentication has been expired and removed from the state.
```
MOZGIII, line 337:
```suggestion
        /// The authentication has been expired and removed from the state.
```

MOZGIII:
Understandable. In the current implementation, can't we correlate the events though? When we see an offense event in the block this pretty much means that one authentication will be taken down... If it still exists by the time the offense is processed, that is.

crates/pallet-bioauth/src/lib.rs
```diff
@@ -526,6 +548,14 @@ pub mod pallet {
             let active_authentications = possibly_expired_authentications
                 .into_iter()
                 .filter(|possibly_expired_authentication| {
+                    if possibly_expired_authentication.expires_at <= current_moment {
```
MOZGIII, line 551: This doesn't look right, I think we should reuse the condition rather than reevaluate it for each authentication. Also, we deliberately avoided sending an event for expiration before to keep the overhead low, maybe we should send just one event with a list of authentications?

dmitrylavrenov:
> Understandable. In the current implementation, can't we correlate the events though? When we see an offense event in the block this pretty much means that one authentication will be taken down... If it still exists by the time the offense is processed, that is.

In the current implementation we can't easy construct the whole way for particular authentication as the authentication can be removed by some other offence than `imOnline`. During the research I got that in case of `grandpa` and `babe` offences we can't get the real reason of removing the authentication.

crates/pallet-bioauth/src/lib.rs
dmitrylavrenov: Thanks, fixed - https://github.com/humanode-network/humanode/pull/649/commits/dc6e74201f31c98ec7abb9869754f3f9fc00fac0

MOZGIII:
I'll have to take some time to think about this implementation. Maybe we want to combine the deauthentication event into one, and provide the details in the fields, rather than having two distinct events...

crates/pallet-bioauth/src/lib.rs
dmitrylavrenov: Reimplemented.

crates/pallet-bioauth/src/lib.rs
```diff
@@ -390,18 +400,29 @@ pub mod pallet {
...
+            });
+            if removed {
+                // Emit an event.
+                Self::deposit_event(Event::AuthenticationsRemoved {
+                    removed: public_keys,
+                    reason,
+                });
+            }
+            removed
         }
```
MOZGIII, line 426: This implementation does not have the ability to accurately communicate whether a public key was actually removed from the set or not. As in, the resulting value also must be an array of public keys instead of being a simple bool.

One of the properties of the old version was that it was possible to invoke `deauthenticate` and get a `false`, which would make sense to log to check where those non-actionable deauthentications come from and how many of them are usually in the chain.

crates/humanode-runtime/src/lib.rs
```diff
@@ -562,19 +565,25 @@ impl
...
                 pallet_humanode_session::Identification::Bootnode(..) => {
                     // Never slash the bootnodes.
                 }
             }
         }
+        if !should_be_deauthenticated.is_empty() {
+            let deauthenticated_public_keys =
+                Bioauth::deauthenticate(should_be_deauthenticated, DeauthenticationReason::Offence);
+            weight = weight.saturating_add(
+                weights.reads_writes(1, u64::from(!deauthenticated_public_keys.is_empty())),
```
MOZGIII, line 584:
```suggestion
                weights.reads_writes(1, deauthenticated_public_keys.len().into()),
```

crates/pallet-bioauth/src/lib.rs
```diff
@@ -390,18 +400,28 @@ pub mod pallet {
                 .any(|authentication| &authentication.public_key == public_key)
         }
 
-        pub fn deauthenticate(public_key: &<T as Config>::ValidatorPublicKey) -> bool {
+        pub fn deauthenticate(
+            public_keys: Vec<<T as Config>::ValidatorPublicKey>,
+            reason: <T as Config>::DeauthenticationReason,
+        ) -> Vec<<T as Config>::ValidatorPublicKey> {
+            let mut removed_public_keys = vec![];
```
MOZGIII, line 407:
```suggestion
            let mut removed_public_keys = Vec::with_capacity(public_keys.len());
```

crates/pallet-bioauth/src/lib.rs
```diff
@@ -523,12 +543,15 @@ pub mod pallet {
...
             let possibly_expired_authentications = <ActiveAuthentications<T>>::get();
             let possibly_expired_authentications_len = possibly_expired_authentications.len();
-            let active_authentications = possibly_expired_authentications
-                .into_iter()
-                .filter(|possibly_expired_authentication| {
-                    possibly_expired_authentication.expires_at > current_moment
-                })
-                .collect::<Vec<_>>();
+            let mut expired_validator_public_keys = vec![];
+            let mut active_authentications = vec![];
```
MOZGIII, line 547:
```suggestion
            let mut expired_validator_public_keys = Vec::with_capacity(possibly_expired_authentications_len);
            let mut active_authentications = Vec::with_capacity(possibly_expired_authentications_len);
```

crates/humanode-runtime/src/lib.rs
```diff
@@ -562,19 +565,25 @@ impl
         }
         let mut weight: Weight = Weight::zero();
         let weights = <Runtime as frame_system::Config>::DbWeight::get();
+        let mut should_be_deauthenticated = vec![];
```
MOZGIII, line 568:
```suggestion
        let mut should_be_deauthenticated = Vec::with_capacity(offenders.len());
```

MOZGIII (approved):
I went over the changes again, and it looks very much improved since I initially checked it. Back then, I liked the idea but was very hesitant about landing the implementation. Very pleased that it is now lacks the shortcoming I've seen initially.
I have also created a follow-up issue #787.

Left a few comments on the technicalities, but it looks good in general.

crates/pallet-bioauth/src/lib.rs
```diff
@@ -523,12 +543,15 @@ pub mod pallet {
...
+            for possible_expired_authentication in possibly_expired_authentications {
+                if possible_expired_authentication.expires_at > current_moment {
+                    active_authentications.push(possible_expired_authentication);
+                } else {
+                    expired_validator_public_keys.push(possible_expired_authentication.public_key)
+                }
+            }
```
MOZGIII, line 554:
```suggestion
            for possibly_expired_authentication in possibly_expired_authentications {
                if possibly_expired_authentication.expires_at > current_moment {
                    // Still active.
                    active_authentications.push(possibly_expired_authentication);
                } else {
                    // Expired!
                    expired_validator_public_keys.push(possibly_expired_authentication.public_key)
                }
            }
```

PR #664, opened 2023-04-22
Switch to workspace dependencies

### LINKED ISSUE #610: Switch to workspace dependencies for substrate and frontier forks
Keeping track of dependency versions/refs in a single location is a substantial improvement to how our versioning mental model is represented in the code. This will result in a great decrease in the number of boring bookkeeping changes we do in our codebase when conducting the updates to substrate/frontier.

MOZGIII:
Closes #610


Cargo.toml
```diff
@@ -2,6 +2,156 @@
...
+fp-self-contained = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+fp-storage = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-base-fee = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-dynamic-fee = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-ethereum = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-evm = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-evm-precompile-modexp = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-evm-precompile-sha3fips = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+pallet-evm-precompile-simple = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
+
```
dmitrylavrenov, line 154: What is the difference including our crates as a part of workspace as well (like author-ext-api,...)? I've seen a lot of projects that include own crates into workspace description as well.

MOZGIII: We can do that too, it is just easier to distinguish between the two groups this way. I don't think it has any meaningful difference.

MOZGIII:
As you can see, CI is red, and the builds are broken. This is without any changes to the `Cargo.lock`, so it must that the features are resolving differently somehow.
I'll investigate this later, but for now, @dmitrylavrenov if you can spot the root cause of this at a glance (i.e. without too much effort) please let me know.

utils/checks/features
```diff
@@ -77,6 +77,10 @@ def list_expected_propagated_features(feature, _package, deps)
     default_enabled = dep[:dependency]['uses_default_features']
     next false if default_enabled && package_feature_enabled_by(dep[:package], 'default', feature)
 
+    # If the package is used as dev dependency - skip it, as we don't need to specify
+    # the feature.
+    next false if dep[:dependency]['kind'] == 'dev'
```
dmitrylavrenov, line 82: To fix - https://github.com/humanode-network/humanode/actions/runs/4786888597/jobs/8511395058?pr=664 as https://github.com/rust-lang/cargo/issues/7754.

MOZGIII:
I see what the problem is. I don't like the idea of having default features disabled for some of the workspace deps, but enabled for the rest. How about we disbale default features for all workspace deps, and the selectively enable them as needed?

MOZGIII:
@dmitrylavrenov thx for the help, I'll take care of the rest.

utils/checks/features
MOZGIII: Right, but it's only when the feature is `std` actually, as we might want to propagate other features properly (i.e. runtime benchmarks and etc).

dmitrylavrenov:
@MOZGIII default features have beet set to false for all deps.
Please, check.

MOZGIII (commented):
I think we must also run local tests for this to be sure (to see if we maybe disabled some features unexpectedly that don't break the compilation, but reduce the effective feature set for instance). Should be alright though.

MOZGIII:
Status update: this can be merged, but we decided to introduce snapshotting of the active feature per dependency, land that at CI so we get an alert when features change, and then ensure this PR doesn't cause a snapshot change to rnsure we only purely rewrite the way we declare the features, and bit the actual sets of features.
@dmitrylavrenov please send an update on the progress of the specifics here once you have a chance

MOZGIII:
See https://github.com/humanode-network/humanode/pull/687

MOZGIII:
#687 is getting merged now, we can continue here

dmitrylavrenov:
@MOZGIII Just merged master and resolved conflict. I guess it can be reviewed now.

MOZGIII:
Can we make it so the features snapshot is unchanged?

MOZGIII:
I suggest that we squash the commits in this PR and rebase on top of `master` so it's easier to achieve the changeset we are looking for here - which is to switch to the workspace dependencies without changes to the features snapshot. This is going to be non-trivial, especially with the current state of this PR, however, it is doable.

dmitrylavrenov:
> I suggest that we squash the commits in this PR and rebase on top of `master` so it's easier to achieve the changeset we are looking for here - which is to switch to the workspace dependencies without changes to the features snapshot. This is going to be non-trivial, especially with the current state of this PR, however, it is doable.

Spent time to properly rebase. Current features snapshot changes are related to `default-features = false` usage based on

```
How about we disbale default features for all workspace deps, and the selectively enable them as needed?
```

dmitrylavrenov:
@MOZGIII Check again, please. Features snapshot hasn't been changed.

MOZGIII (commented):
Looks great! I'll take some time to review to make sure I understand the tradeoffs taken here.

MOZGIII:
It would be nice to add a clippy lint to only allow workspace dependencies. This doesn't look too difficult, check out this for a reference: https://github.com/rust-lang/rust-clippy/tree/master/clippy_lints/src/cargo

PR #679, opened 2023-06-29
Implement `pallet-currency-swap` logic

### LINKED ISSUE #678: Implement currency swap logic
Check details - https://github.com/humanode-network/arch/pull/49

dmitrylavrenov:
Closes #678


crates/pallet-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        /// Swap balances.
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::swap())]
+        pub fn swap(
+            origin: OriginFor<T>,
+            account_id_to: T::AccountIdTo,
+            balance: BalanceFromOf<T>,
+        ) -> DispatchResult {
```
MOZGIII, line 73: I think we should do this within a storage layer - so the swap is atomic

crates/pallet-currency-swap/src/traits.rs
```diff
@@ -0,0 +1,20 @@
+//! Traits we use and expose.
+
+use frame_support::{sp_runtime::DispatchError, traits::Currency};
+
+/// Currency swap interface.
+pub trait CurrencySwap<AccountIdFrom, AccountIdTo> {
```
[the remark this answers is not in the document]
dmitrylavrenov: Done

crates/pallet-currency-swap/src/traits.rs
```diff
@@ -0,0 +1,20 @@
...
+    type To: Currency<AccountIdTo>;
+    /// An error happens during the actual balances swap.
+    type Error: Into<DispatchError>;
+
+    /// Swap balances.
+    fn swap(
+        account_id_from: &AccountIdFrom,
+        account_id_to: &AccountIdTo,
+        amount: <Self::From as Currency<AccountIdFrom>>::Balance,
+    ) -> Result<<Self::To as Currency<AccountIdTo>>::Balance, Self::Error>;
```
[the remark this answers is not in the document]
dmitrylavrenov: Done

crates/pallet-currency-swap/src/lib.rs
dmitrylavrenov: Done

MOZGIII (approved):

MOZGIII:
Let's add benches too

crates/pallet-currency-swap/src/benchmarking.rs
```diff
@@ -0,0 +1,123 @@
...
+use frame_support::{assert_ok, dispatch::DispatchResult, traits::Currency};
+use frame_system::RawOrigin;
+
+use crate::*;
+
+const WITHDRAW_BALANCE: u32 = 100;
+const DEPOSIT_BALANCE: u32 = 100;
+
+/// The benchmarking extension for the currency swap interface.
+pub trait CurrencySwap<AccountIdFrom, AccountIdTo>:
```
MOZGIII, line 13: I think this should be a part of the `Interface`, not a trait extension of the `primitives_currency_swap::CurrencySwap` - because think of how you would use it, you'd have to make a newtype just to implement the trait at runtime, no?

crates/pallet-currency-swap/src/benchmarking.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn to() -> <Self as Config>::AccountIdTo {
+        use sp_std::str::FromStr;
+
+        mock::EvmAccountId::from_str("1000000000000000000000000000000000000001").unwrap()
+    }
+}
+
+#[cfg(test)]
+impl CurrencySwap<mock::AccountId, mock::EvmAccountId>
+    for <crate::mock::Test as super::Config>::CurrencySwap
```
MOZGIII, line 95: Refer to the type explicitly here

crates/pallet-currency-swap/src/benchmarking.rs
MOZGIII: Gave this a more-in depth look - for sure we need to move this functionality into the `Interface`. The point of an `Interface` is that you implement it for the `Runtime`, and thus guarantee there are no conflicting implementations possible since it is a single-purpose type.

If we wanted to use this in runtime, we'd have to implement `pallet_currency_swap::benchmarking:: CurrencySwap` for `bridge_currency_swap::OneToOne`, at runtime crate, which are both foreign to that crate and it won't work in Rust.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/679/commits/ab585b22dbebdb23315ea3de80ab2f0fe90666ef

crates/pallet-currency-swap/src/benchmarking.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/679/commits/ab585b22dbebdb23315ea3de80ab2f0fe90666ef

crates/pallet-currency-swap/src/benchmarking.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn from() -> <Self as frame_system::Config>::AccountId;
+
+    /// Obtain withdraw balances amount.
+    fn withdraw_amount() -> FromBalanceOf<Self>;
+
+    /// Obtain an account id balances deposited to.
+    fn to() -> <Self as Config>::AccountIdTo;
+
+    /// Obtain deposit balances amount.
+    fn deposit_amount() -> ToBalanceOf<Self>;
```
MOZGIII, line 30:
```suggestion
    /// Obtain an account id balances withdrawed from.
    fn from_account_id() -> <Self as frame_system::Config>::AccountId;

    /// Obtain withdraw balances amount.
    fn from_balance() -> FromBalanceOf<Self>;

    /// Obtain an account id balances deposited to.
    fn to_account_id() -> <Self as Config>::AccountIdTo;

    /// Obtain deposit balances amount.
    fn to_balance() -> ToBalanceOf<Self>;
```

crates/pallet-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+                    WithdrawReasons::TRANSFER,
+                    ExistenceRequirement::AllowDeath,
+                )?;
+                let withdrawed_amount = withdrawed_imbalance.peek();
+
+                let deposited_imbalance =
+                    T::CurrencySwap::swap(withdrawed_imbalance).map_err(|error| {
+                        // Here we undo the withdrawl to avoid having a dangling imbalance.
+                        FromCurrencyOf::<T>::resolve_creating(&who, error.incoming_imbalance);
+                        error.cause.into()
```
MOZGIII, line 123:
```suggestion
                        error.cause
```
Needless into

crates/pallet-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+            deposited_amount: ToBalanceOf<T>,
+        },
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        /// Swap balances.
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::swap())]
+        pub fn swap(
```
MOZGIII, line 103: We might want to add a `swap_keep_alive` call here, to expose a `ExistenceRequirement::KeepAlive` variant of this call.

MOZGIII (approved):

crates/pallet-currency-swap/src/lib.rs
dmitrylavrenov: It requires, already tried.

crates/pallet-currency-swap/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/issues/689

PR #682, opened 2023-06-29
Currency swap precompile

MOZGIII:
To do:

- [x] implementation
- [x] tests

### DIFF
--- a/crates/precompile-currency-swap/Cargo.toml
+++ b/crates/precompile-currency-swap/Cargo.toml
@@ -22,11 +22,12 @@ sp-runtime = { default-features = false, git = "https://github.com/humanode-netw
 sp-std = { default-features = false, git = "https://github.com/humanode-network/substrate", branch = "locked/polkadot-v0.9.38" }
 
 [dev-dependencies]
+hex-literal = "0.4"
 mockall = "0.11"
-pallet-evm = { default-features = false, git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
-pallet-evm-system = { default-features = false, git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
-pallet-evm-balances = { default-features = false, git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
 pallet-balances = { default-features = false, git = "https://github.com/humanode-network/substrate", branch = "locked/polkadot-v0.9.38" }
+pallet-evm = { default-features = false, git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38" }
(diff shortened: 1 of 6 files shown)


crates/precompile-currency-swap/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+[dependencies]
+precompile-utils = { path = "../precompile-utils", default-features = false }
+primitives-currency-swap = { path = "../primitives-currency-swap", default-features = false }
+
+codec = { package = "parity-scale-codec", version = "3.2.2", default-features = false, features = ["derive"] }
+fp-evm = { git = "https://github.com/humanode-network/frontier", branch = "locked/polkadot-v0.9.38", default-features = false }
+frame-support = { default-features = false, git = "https://github.com/humanode-network/substrate", branch = "locked/polkadot-v0.9.38" }
+frame-system = { default-features = false, git = "https://github.com/humanode-network/substrate", branch = "locked/polkadot-v0.9.38" }
+getrandom = { version = "0.2", features = ["js"] }
+num_enum = { version = "0.6.0", default-features = false }
```
dmitrylavrenov, line 16: Do we use these dependencies somewhere?

crates/precompile-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,778 @@
...
+#[test]
+fn swap_works() {
+    new_test_ext().execute_with_ext(|_| {
+        let alice_evm = H160::from(hex_literal::hex!(
+            "1000000000000000000000000000000000000001"
+        ));
+        let alice = AccountId::from(hex_literal::hex!(
+            "1000000000000000000000000000000000000000000000000000000000000001"
+        ));
+        let alice_balance = 100 * 10u128.pow(18);
```
dmitrylavrenov, line 25:
```suggestion
        let alice_evm_balance = 100 * 10u128.pow(18);
```

dmitrylavrenov (approved):

crates/precompile-currency-swap/Cargo.toml
MOZGIII: We might not, I'll have to check the list as I used our other precompile as a template

crates/precompile-currency-swap/src/tests.rs
MOZGIII: [`dc1ff81` (#682)](https://github.com/humanode-network/humanode/pull/682/commits/dc1ff81313e3995b220eb0c4856831456c54fdd7)

crates/precompile-currency-swap/Cargo.toml
MOZGIII: Reduced the set of the dependencies manually to the amount that I found works (might be able to reduce further via reexports that I missed, but I figured it's good enough).

The `num_enum` is used implicitly by the precompile utils macro that generates selectors from the enum. `getrandom` was not used.

PR #685, opened 2023-06-30
Automatic currency-swap proxy primitives

MOZGIII:


MOZGIII:
@dmitrylavrenov, so check this out. This is what will allow us to pass the fees and dust removals to the native treasury and fees pots.
One issue I have discovered while working on this is our `CurrencySwap::swap` call does not allow access to the original `NegativeImbalance` in case of failure. We should probably rework the `CurrencySwap` trait API and implementation such that the API ensures we don't ever lose the `NegativeImbalance`  - as it should never just disappear in the swap.
Please let me know what you think.
If you agree - let's get to work! We can merge all we have currently to `master` and then apply a corrective fix over all of the implementations that involve `CurrencySwap`.

MOZGIII:
Change of plans, we merge the fix early and apply updates to the respective PRs.

MOZGIII:
@dmitrylavrenov this has been rebased and finalized for the merge, please take another look for a final review

dmitrylavrenov (approved):

PR #687, opened 2023-07-01
Features snapshot check

MOZGIII:


dmitrylavrenov (approved):

MOZGIII:
We'll merge this after the EVM work is done to avoid the possible disruption

MOZGIII:
@dmitrylavrenov ok, so, this works, but it completely breaks dependabot autoupdates. This is why I am a bit reluctant to merge it. Is it worth it, what do you think?

PR #694, opened 2023-07-04
Check existential deposit when creating an account during swap

### LINKED ISSUE #692: Check existential deposit when creating an account during swap
We have discovered an issue with account creation in both `pallet_currency_swap` and `precompile_currency_swap` in that they can both lead to the loss of funds when the amount swapped is smaller that the `To` currency existential deposit.

Currently, the account gets dusted before it is created.
We should change is the following ways:
1. Use `try_mutate_account` if possible (maybe we can't though).
2. Handle the case where the value swapped is less than the existential deposit of the currency, and fail the swap operation. In substrate this is done usually at the pallet call level, so we should do the same: add the check to the `pallet_currency_swap` and `precompile_currency_swap` calls.

dmitrylavrenov:
Closes #692


crates/primitives-currency-swap/src/lib.rs
```diff
@@ -3,15 +3,20 @@
...
+    sp_runtime::DispatchError,
+    traits::{fungible::Inspect, Currency},
+};
 
 /// Currency swap interface.
 pub trait CurrencySwap<AccountIdFrom, AccountIdTo> {
     /// The currency to convert from.
-    type From: Currency<AccountIdFrom>;
+    type From: Currency<AccountIdFrom>
+        + Inspect<AccountIdFrom, Balance = <Self::From as Currency<AccountIdFrom>>::Balance>;
```
MOZGIII, line 15: Why?

dmitrylavrenov: To use `can_deposit`. `Balances` and `Evmbalances` implements it.

MOZGIII (commented):
Actually, it seems we don't need to do any checks on the calls:

https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs#L284-L300

I am pretty sure the currency is supposed to handle that:

https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs#L1501-L1563
https://github.com/humanode-network/frontier/blob/locked/polkadot-v0.9.38/frame/evm-balances/src/lib.rs#L439-L497

The question is why doesn't it work as expected?

dmitrylavrenov:
> Actually, it seems we don't need to do any checks on the calls:
> 
> https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs#L284-L300
> 
> I am pretty sure the currency is supposed to handle that:
> 
> https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs#L1501-L1563 https://github.com/humanode-network/frontier/blob/locked/polkadot-v0.9.38/frame/evm-balances/src/lib.rs#L439-L497
> 
> The question is why doesn't it work as expected?

As we don't use the actual transfer, we use default `resolve_creating`.

crates/primitives-currency-swap/src/lib.rs
MOZGIII: I see, the question is why do we need a manual extra check if we are about to actually do the deposit in the same flow

MOZGIII:
> > Actually, it seems we don't need to do any checks on the calls:
> > [humanode-network/substrate@`master`/frame/balances/src/lib.rs#L284-L300](https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs?rgh-link-date=2023-07-04T13%3A41%3A01Z#L284-L300)
> > I am pretty sure the currency is supposed to handle that:
> > [humanode-network/substrate@`master`/frame/balances/src/lib.rs#L1501-L1563](https://github.com/humanode-network/substrate/blob/master/frame/balances/src/lib.rs?rgh-link-date=2023-07-04T13%3A41%3A01Z#L1501-L1563) [humanode-network/frontier@`locked`/polkadot-v0.9.38/frame/evm-balances/src/lib.rs#L439-L497](https://github.com/humanode-network/frontier/blob/locked/polkadot-v0.9.38/frame/evm-balances/src/lib.rs?rgh-link-date=2023-07-04T13%3A41%3A01Z#L439-L497)
> > The question is why doesn't it work as expected?
> 
> As we don't use the actual transfer, we use default `resolve_creating`.

Ah, got it! This must be the reason it doesn't do those checks. Ok then!

MOZGIII (approved):
Looks good so far

dmitrylavrenov:
> Looks good so far

Will add some tests later.

crates/bridge-pot-currency-swap/src/lib.rs
MOZGIII, line 1: We should add the tests for the bridge as well... Maybe separately.

MOZGIII:
But then, if we follow the trail of calls, we invoke the `resolve_creating`, which calls `deposit_creating`, which, in turn, does have the check: https://github.com/humanode-network/frontier/blob/07429a3b559c7277fa20024587f67ae809556054/frame/evm-balances/src/lib.rs#L608C36-L608C36

MOZGIII (changes requested):
Hold on, we need to figure out exactly what's going on

MOZGIII (approved):
I think we can merge this for now, and clean up this later once we have switched to `fungible` traits.

MOZGIII (approved):

Issue #695, opened 2023-07-04
Docs: native-to-EVM bridge math


MOZGIII:
Moved to internal docs

PR #698, opened 2023-07-19
Proper initialization of the bridge pot account balances in genesis

dmitrylavrenov:
- [x] Bridge pot currency swap logic as substrate pallet
- [x] Integrate changes into runtime and tests
- [x] Introduce genesis verifier
- [x] ~Use https://github.com/humanode-network/humanode/issues/695 at genesis verifier~ - https://github.com/humanode-network/humanode/issues/699


crates/pallet-bridge-pot-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+                T::GenesisVerifier::verify(),
+                "invalid genesis bridge pot currency swap related data"
+            );
+        }
+    }
+}
+
+/// A [`primitives_currency_swap::CurrencySwap`] implementation that does the swap using two
+/// "pot" accounts for each of end swapped currencies.
+pub struct CurrencySwap<Pallet, ExistenceRequirement>(PhantomData<(Pallet, ExistenceRequirement)>);
```
MOZGIII, line 126:
```suggestion
pub struct CurrencySwap<Config, ExistenceRequirement>(PhantomData<(Config, ExistenceRequirement)>);
```

crates/pallet-bridge-pot-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+
+/// The interface to verify bridge pot currency swap related data at genesis.
+pub trait GenesisVerifier {
```
MOZGIII, line 28: I believe this crate should also offer a verifier implementation, and that this implementation should be properly tested, rather then mock-tested.

MOZGIII (commented):
This is great! Left a few comments, and we probably need to add the verifier itself... Which could be added in a separate PR.
This looks good, so probably would be a good time to merge and continue work in the subsequent PRs...

Just one question though - we have two of these pallets, would we also have two validators? It makes sense, kind of, but I'm just asking...

dmitrylavrenov:
> This is great! Left a few comments, and we probably need to add the verifier itself... Which could be added in a separate PR. This looks good, so probably would be a good time to merge and continue work in the subsequent PRs...
> 
> Just one question though - we have two of these pallets, would we also have two validators? It makes sense, kind of, but I'm just asking...

I see, it's great questions as I was thinking about it as well. In such implementation we have 2 verifiers for each pallet. I was a little bit confused as it has some pros and cons. But during tests implementation I've come up that it can provide more independent features, imo.

crates/pallet-bridge-pot-currency-swap/src/lib.rs
dmitrylavrenov: I've created an issue - https://github.com/humanode-network/humanode/issues/699 to implement it in separate PR based on your review comment.

MOZGIII (approved):

PR #700, opened 2023-07-24
Add the actual verifier implementation into `pallet-bridge-pot-currency-swap`

### LINKED ISSUE #699: Add the actual verifier implementation into `pallet-bridge-pot-currency-swap`

dmitrylavrenov:
Closes #699


crates/humanode-peer/src/chain_spec.rs
```diff
@@ -344,7 +356,15 @@ fn testnet_genesis(
...
-                    evm_genesis_account(INITIAL_POT_ACCOUNT_BALANCE),
+                    evm_genesis_account(
+                        DEV_ACCOUNT_BALANCE
+                            .checked_mul(endowed_accounts.len().try_into().unwrap()).unwrap()
+                            .checked_add(INITIAL_POT_ACCOUNT_BALANCE.checked_mul(pot_accounts.len().try_into().unwrap()).unwrap()
+                            .checked_add(
+                                // Own bridge pot minimum balance.
+                                <humanode_runtime::EvmBalances as frame_support::traits::Currency<EvmAccountId>>::minimum_balance(),
+                            ).unwrap()).unwrap()
+                    ),
```
MOZGIII, line 367: Can we place this calculation in a separate fn and make it more explicit and testable? The same goes for the formula for the native account bridge

dmitrylavrenov: Got it, sure. But what do you mean in term of `testable`? As it's tested in `chain_spec` tests.

MOZGIII: The fn itself could have a number of unit tests focusing on illustrating why that fn works right.

dmitrylavrenov: Redesigned in a little bit different way. Please, check.

MOZGIII: Well, this does not help with the growing complexity at `testnet_genesis` fn.

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -262,6 +262,33 @@ fn testnet_genesis(
     robonode_public_key: robonode::PublicKey,
     bootnodes: Vec<AccountId>,
 ) -> GenesisConfig {
+    let pot_accounts = vec![
```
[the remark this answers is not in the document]
dmitrylavrenov: Renamed.

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/6daee15de13c754f806225d468c20cf60848edc5

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -252,6 +252,81 @@ const EXISTENTIAL_DEPOSIT: Balance = 500;
 /// The initial pot accounts balance for testnet genesis.
 const INITIAL_POT_ACCOUNT_BALANCE: Balance = EXISTENTIAL_DEPOSIT + DEV_ACCOUNT_BALANCE;
 
+/// A helper function to prepare native and evm accounts.
+#[allow(clippy::type_complexity)]
+fn prepare_accounts_balances(
```
MOZGIII, line 257: I propose two changes here:
- this fn should take two lists of account/balance pair instead of taking four lists of accouts;
- this fn should return simple balances; the conversion to `fp_evm::GenesisAccount`s for instance should be kept outside of this fn
- instead of returning the whole list of accounts, this fn could return just a tuple of two balances - for native and evm bridge

What do you think?

dmitrylavrenov: The first suggestions sounds helpful to avoid focusing on different balance values for different account types. The second and the third ones are outcome of the first one :)

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/48e26183a49c09e248689abbc1de043d30359e80

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -342,17 +384,15 @@ fn testnet_genesis(
...
             accounts: {
-                let evm_pot_accounts = vec![(
-                    humanode_runtime::EvmToNativeSwapBridgePot::account_id(),
-                    evm_genesis_account(INITIAL_POT_ACCOUNT_BALANCE),
-                )];
-
-                evm_pot_accounts
+                basic_evm_accounts
                     .into_iter()
+                    .map(|k| (k.0, evm_genesis_account(k.1)))
```
MOZGIII, line 389:
```suggestion
                    .map(|(account_id, balance)| (account_id, evm_genesis_account(balance)))
```

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -252,6 +252,35 @@ const EXISTENTIAL_DEPOSIT: Balance = 500;
 /// The initial pot accounts balance for testnet genesis.
 const INITIAL_POT_ACCOUNT_BALANCE: Balance = EXISTENTIAL_DEPOSIT + DEV_ACCOUNT_BALANCE;
 
+/// A helper function to calculate bridge pot balances.
+#[allow(clippy::type_complexity)]
```
MOZGIII, line 256: I don't like this, definitely not justified

crates/pallet-bridge-pot-currency-swap/src/genesis_verifier.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// The implementation that requires bridge pot balance values be balanced.
+pub struct Balanced<Pallet>(PhantomData<Pallet>);
+
+impl<T: Config<I>, I: 'static> Balanced<Pallet<T, I>> {
+    /// A function to calculate expected [`Config::PotTo`] balance based on the provided list of
+    /// all [`Config::AccountIdFrom`] balances except [`Config::PotFrom`] balance.
+    pub fn calculate_expected_to_bridge_balance(
+        from_balances: &[<T::CurrencyFrom as Currency<T::AccountIdFrom>>::Balance],
+    ) -> Result<<T::CurrencyTo as Currency<T::AccountIdTo>>::Balance, DispatchError> {
```
MOZGIII, line 22:
```suggestion
    pub fn calculate_expected_to_bridge_balance<Iter: IntoIter<Item = <T::CurrencyFrom as Currency<T::AccountIdFrom>>::Balance>>(
        from_balances: Iter,
    ) -> Result<<T::CurrencyTo as Currency<T::AccountIdTo>>::Balance, DispatchError> {
```

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -342,17 +384,15 @@ fn testnet_genesis(
...
-                let evm_pot_accounts = vec![(
-                    humanode_runtime::EvmToNativeSwapBridgePot::account_id(),
-                    evm_genesis_account(INITIAL_POT_ACCOUNT_BALANCE),
-                )];
-
-                evm_pot_accounts
+                basic_evm_accounts
                     .into_iter()
+                    .map(|(account_id, balance)| (account_id, evm_genesis_account(balance)))
                     .chain(
```
MOZGIII, line 390: If you swap .chain and .map order you should be able to get rid of one `evm_genesis_account` call literal

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -252,6 +252,35 @@ const EXISTENTIAL_DEPOSIT: Balance = 500;
 /// The initial pot accounts balance for testnet genesis.
 const INITIAL_POT_ACCOUNT_BALANCE: Balance = EXISTENTIAL_DEPOSIT + DEV_ACCOUNT_BALANCE;
 
+/// A helper function to calculate bridge pot balances.
+#[allow(clippy::type_complexity)]
+fn get_bridge_pot_balances(
+    basic_native_accounts: &[(AccountId, Balance)],
+    basic_evm_accounts: &[(EvmAccountId, Balance)],
+) -> (Balance, Balance) {
```
MOZGIII, line 260: If we can take two `impl IntoIter` here instead of slices we can remove `.collect`s and thus eliminate extra iterations

crates/pallet-bridge-pot-currency-swap/src/genesis_verifier.rs
MOZGIII: Do we have to use a `DispatchError` here? This is not a call, and we could have a more precise error here...

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: Forgot to remove after refactoring, thanks.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/d58e2871599e3dbeffb7cbdb1363acdca0ba8e39

crates/pallet-bridge-pot-currency-swap/src/genesis_verifier.rs
dmitrylavrenov: Great suggestion - https://github.com/humanode-network/humanode/pull/700/commits/14909bec64702376f433530bc54726aea4a1b05b

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/14e6cd096a332224ee79fbe21365b8da22b57154

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: My bad, thanks - https://github.com/humanode-network/humanode/pull/700/commits/14e6cd096a332224ee79fbe21365b8da22b57154.

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -252,6 +252,28 @@ const EXISTENTIAL_DEPOSIT: Balance = 500;
...
 
+/// A helper function to calculate bridge pot balances.
+fn get_bridge_pot_balances(
+    basic_native_accounts: impl IntoIterator<Item = (AccountId, Balance)>,
+    basic_evm_accounts: impl IntoIterator<Item = (EvmAccountId, Balance)>,
+) -> (Balance, Balance) {
+    let native_bridge_pot_balance = pallet_bridge_pot_currency_swap::genesis_verifier::Balanced::<
+        humanode_runtime::EvmToNativeSwapBridge,
+    >::calculate_expected_to_bridge_balance(
+        basic_evm_accounts.into_iter().map(|acc| acc.1)
```
MOZGIII, line 263:
```suggestion
        basic_evm_accounts.into_iter().map(|(_, balance)| balance)
```

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -262,6 +284,35 @@ fn testnet_genesis(
...
+            .into_iter(),
+        )
+        .collect::<Vec<_>>();
+    let basic_evm_accounts = evm_endowed_accounts
+        .into_iter()
+        .map(|k| (k, DEV_ACCOUNT_BALANCE))
+        .collect::<Vec<_>>();
+
+    let (native_bridge_pot_balance, evm_bridge_pot_balance) =
+        get_bridge_pot_balances(basic_native_accounts.clone(), basic_evm_accounts.clone());
```
MOZGIII, line 314:
```suggestion
        get_bridge_pot_balances(basic_native_accounts.iter().copied(), basic_evm_accounts.iter().copied());
```

crates/pallet-bridge-pot-currency-swap/src/genesis_verifier.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    /// The arithmetic operation error.
+    Arithmetic(ArithmeticError),
+}
+
+impl<T: Config<I>, I: 'static> Balanced<Pallet<T, I>> {
+    /// A function to calculate expected [`Config::PotTo`] balance based on the provided list of
+    /// all [`Config::AccountIdFrom`] balances except [`Config::PotFrom`] balance.
+    pub fn calculate_expected_to_bridge_balance(
```
MOZGIII, line 27: I'd like this logic tested with chain configurations with non-equal `Currency::minimum_balance` values. The rationale for this is here I see that `CurrencyTo::minimum_balance()` is added.

Just as heads up, I would like to see the exact numbers hardcoded at the expectations in the tests, rather than them being computed as well.

For instance, do `500 /* ed from */ + 123 /* ed to */` instead of `CurrencyTo::minimum_balance() + CurrencyFrom::minimum_balance()` at the expectations.

The rationale for this is it would be a more explicit test that way, that would ensure that not only the calculations are tested, but also the underlying assumptions are matched against the effective calculations, and will help us keep an eye on this logic in the context of the possible future changes. This rationale could as well be added at the tests...

---

Adding these tests could be done just for this fn as the units, or for the whole runtime - whichever you find more fitting.

MOZGIII, line 27:
```suggestion
    pub fn calculate_expected_bridge_balance(
```

Although specific, the `to` bit here is confusing. I propose we drop it, as it is clear from the types which bridge the balance is for.

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: `sp_runtime::AccountId` doesn't implement `Copy`.

crates/pallet-bridge-pot-currency-swap/src/genesis_verifier.rs
dmitrylavrenov: Changed the mocked runtime and test with different existential deposit values - https://github.com/humanode-network/humanode/pull/700/commits/ade7c56c6c6ae24d02a0211a04e8ff1186b833d5

crates/pallet-bridge-pot-currency-swap/src/tests.rs
```diff
@@ -1,24 +1,37 @@
...
 #[should_panic = "invalid genesis bridge pot currency swap related data"]
 fn genesis_verifier_false() {
     with_runtime_lock(|| {
-        let verify_ctx = MockGenesisVerifier::verify_context();
-        verify_ctx.expect().once().return_const(false);
+        let verify_lr_ctx = MockGenesisVerifierLR::verify_context();
+        let verify_rl_ctx = MockGenesisVerifierLR::verify_context();
+        verify_lr_ctx.expect().once().return_const(false);
+        // Swap bridge left genesis config presents early than swap bridge right genesis config
+        // at mocked runtime.
```
MOZGIII, line 18: What?

crates/pallet-bridge-pot-currency-swap/src/tests.rs
MOZGIII, line 1: We now need a test that would demonstrate the coherence of `calculate_expected_bridge_balance` with the `GenesisVerifier` when the EDs are different. I.e. no mocking.

---

Let's also try doing the swaps in the tests - just to be sure we can actually swap the funds as expected.
Ensure that the whole swappable balance can be swapped in one direction, and then in the other direction.

crates/humanode-peer/src/chain_spec.rs
dmitrylavrenov: `cloned`

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/c44caa303b18e6b6ea6a9c48ce8c857b1a059d9f

crates/pallet-bridge-pot-currency-swap/src/tests.rs
MOZGIII: Well, we could invoke the swap itself, right? Why would we want to test them at this level - maybe do it at runtime with the actual network configuration?

crates/pallet-bridge-pot-currency-swap/src/tests.rs
```diff
@@ -1,44 +1,135 @@
...
+use crate::{Balanced, BalancedError};
 
-/// This test verifies that the genesis builder correctly ensures genesis verifier result in case
-/// it returns false.
+/// This test verifies that the genesis builder correctly ensures genesis bridge pot balances
+/// values that are balanced.
 #[test]
-#[should_panic = "invalid genesis bridge pot currency swap related data"]
-fn genesis_verifier_false() {
+fn genesis_balanced_true() {
```
MOZGIII, line 9: Add another test like this but with only bridge pot accounts (i.e. no other accounts)

crates/pallet-bridge-pot-currency-swap/src/tests.rs
```diff
@@ -1,44 +1,135 @@
...
+                    (4215, 100),
+                    (SwapBridgeRightPot::account_id(), 120),
+                ],
+            },
+            swap_bridge_left_pot: pallet_pot::GenesisConfig {
+                initial_state: pallet_pot::InitialState::Unchecked,
+            },
+            swap_bridge_right_pot: pallet_pot::GenesisConfig {
                 initial_state: pallet_pot::InitialState::Unchecked,
             },
```
MOZGIII, line 36: Why unchecked? This should be `Initialized`

crates/pallet-bridge-pot-currency-swap/src/balances_values.rs
```diff
@@ -0,0 +1,78 @@
...
+        ArithmeticError,
+    },
+    sp_std::marker::PhantomData,
+    traits::{Currency, Get},
+};
+
+use super::Config;
+
+/// The implementation that requires bridge pot balances values be balanced.
+pub struct Balanced<T: Config<I>, I: 'static>(PhantomData<(T, I)>);
```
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/4513b1b608ae8a21188869e8d1b893bd30166b6b

crates/pallet-bridge-pot-currency-swap/src/tests.rs
dmitrylavrenov: My bad, thanks, fixed - https://github.com/humanode-network/humanode/pull/700/commits/ba986051de0010f441215de9a226ae9c3c9407e9.

crates/pallet-bridge-pot-currency-swap/src/tests.rs
dmitrylavrenov: Added

crates/pallet-bridge-pot-currency-swap/src/tests.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/700/commits/85bbf09854f0600f64407003dfe2639d764848bd

dmitrylavrenov:
In favour of https://github.com/humanode-network/humanode/pull/716

dmitrylavrenov:
Closes in favour of https://github.com/humanode-network/humanode/pull/716

PR #702, opened 2023-07-24
Fix `WeightPerGas` for `pallet-evm`

dmitrylavrenov:


MOZGIII (approved):

MOZGIII:
This is a great hotfix, but we still need to revisit this. Even if the numbers are good - they should probably be in the constants module. Also, we might be able to take the `WEIGHT_MILLISECS_PER_BLOCK` instead of hard-coding it.
Overall, this needs another pass - but it can be done later.

PR #713, opened 2023-08-04
Tweaks to the developer experience with EVM

MOZGIII:
This PR has a few changes:
- Renames the CLI commands from `ethereum` to `evm`
- Switches the dev and testnet genesis to using standard Ethereum BIP44 account generation logic with the standard substrate dev seed
- Refactors the code to share the implementation of the chainspec and cli commands account generation/visualization.

Overall, this makes it easier to use our node with `hardhat`, as there are now usable pre-provisioned accounts that can be used for running the tests.

---

The old way of generating an EVM account is actually broken with respect to how usable it is with various wallets.


crates/crypto-utils-evm/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn from_mnemonic_bip39(
+        mnemonic: &bip39::Mnemonic,
+        password: &str,
+        derivation_path: &bip32::DerivationPath,
+    ) -> Self {
+        // Retrieve the seed from the mnemonic.
+        let seed = bip39::Seed::new(mnemonic, password);
+
+        // Derives the private key from.
+        let ext = bip32::XPrv::derive_from_path(seed, derivation_path).unwrap();
```
dmitrylavrenov, line 29: Are we going just doing `unwrap` without returning any error ?

MOZGIII: Not really, we need to do something about it; also, a crate and unused vars cleanup is pending.

dmitrylavrenov (approved):
The changes are clear and useful!

crates/crypto-utils-evm/src/lib.rs
MOZGIII: Should be ready now, please take another look

dmitrylavrenov: One `unwrap` was missed. Everything is fine. Thanks.

crates/crypto-utils-evm/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /// Construct the key info from the BIP39 mnemonic using BIP44 convenions.
+    pub fn from_mnemonic_bip44(
+        mnemonic: &bip39::Mnemonic,
+        password: &str,
+        account_index: Option<u32>,
+    ) -> Result<Self, FromMnemonicBip39Error> {
+        let derivation_path = format!("m/44'/60'/0'/0/{}", account_index.unwrap_or(0));
+        let derivation_path = derivation_path.parse().unwrap();
```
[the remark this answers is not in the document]
MOZGIII: Yep, fixed it [`7404370` (#713)](https://github.com/humanode-network/humanode/pull/713/commits/740437006edf3af05b906cbaa62c32c9bedd4b50)

PR #716, opened 2023-08-07
Implement bridges pot currency swap accounts inititialization

dmitrylavrenov:


crates/pallet-bridges-initializer-currency-swap/src/mock.rs
```diff
@@ -0,0 +1,206 @@
...
+impl BalanceMaker<AccountId, EvmBalances> for EvmBridgeBalanceMaker {
+    const IS_SWAPPABLE_CHANGED: bool = false;
+
+    fn make_balance(amount: Balance) -> Result<(), DispatchError> {
+        let imbalance = EvmBalances::issue(amount);
+
+        EvmBalances::resolve_creating(&SwapBridgeNativeToEvmPotAccountId::get(), imbalance);
+        Ok(())
+    }
+}
```
MOZGIII, line 169: Let's drop these and just inline the necessary operations. It would just be simpler, and less knobs to test

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+};
+pub use pallet::*;
+
+#[cfg(test)]
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+pub trait BalanceMaker<AccountId, C: Currency<AccountId>> {
```
MOZGIII, line 21: See https://github.com/humanode-network/humanode/pull/716/files#r1286503310

crates/pallet-bridges-initializer-currency-swap/src/mock.rs
dmitrylavrenov: Dropped - https://github.com/humanode-network/humanode/pull/716/commits/8c0c96be4f200560e1ea9fb782a324213b54689c

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/8c0c96be4f200560e1ea9fb782a324213b54689c

crates/pallet-bridges-initializer-currency-swap/src/upgrade_init.rs
```diff
@@ -0,0 +1,33 @@
+//! Initialization of the bridge pot accounts on runtime upgrade.
+
+use frame_support::pallet_prelude::*;
+
+use crate::Pallet;
+
+/// Initialize the bridges pot accounts.
+pub fn on_runtime_upgrade<T: crate::Config>() -> Weight {
+    match Pallet::<T>::initialize() {
```
MOZGIII, line 9: Here we need to check some sort of condition and only perform the init if necessary. I suggest that we add an explicit boolean (`has_initialized`) to the state to make things easier.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/a80bb29f3e19794ce39bbe7b91bc037249604db6

crates/pallet-bridges-initializer-currency-swap/src/mock/v1.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// This function basically just builds a genesis storage key/value store according to
+// our desired mockup.
+pub fn new_test_ext_with(genesis_config: GenesisConfig) -> sp_io::TestExternalities {
+    let storage = genesis_config.build_storage().unwrap();
+    storage.into()
+}
+
+pub fn runtime_lock() -> std::sync::MutexGuard<'static, ()> {
```
MOZGIII, line 110: This runtime lock should probably be shared across the versions - as the scope of it is the access to a global state, not a particular version's state.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,238 @@
...
+                initial_state: pallet_pot::InitialState::Initialized,
+            },
+            ..Default::default()
+        };
+        new_test_ext_with(config).execute_with(move || {});
+    })
+}
+
+#[test]
+fn runtime_upgrade() {
```
MOZGIII, line 191: Good progress here! There are a couple of issues though; first - I can't tell where the v1 vs v2 is.
Second - it seems like with this approach the test externalities should be kind of shared across both versions, like like the runtime lock.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
MOZGIII, line 1: Let's add a struct type for address/balance pair, so we can avoid using tuple access (i.e. `.1`). This is a small change but will hugely improve the readability.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,238 @@
+use frame_support::traits::{Currency, OnRuntimeUpgrade};
+
+use crate::mock::{v1, v2::*};
```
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/421ef86442344cb25ff3c54d8fd9898e50afad1f

crates/pallet-bridges-initializer-currency-swap/src/mock/v1.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/55ad986186808aad0dc49816ce9b07638f84329e

crates/pallet-bridges-initializer-currency-swap/src/upgrade_init.rs
```diff
@@ -0,0 +1,41 @@
+//! Initialization of the bridge pot accounts on runtime upgrade.
+
+use frame_support::pallet_prelude::*;
+
+use crate::{Config, Pallet};
+
+/// Initialize the bridges pot accounts.
+pub fn on_runtime_upgrade<T: Config>() -> Weight {
+    let is_balanced = Pallet::<T>::is_balanced().unwrap_or_default();
```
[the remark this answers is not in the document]
dmitrylavrenov: Currency related data is something that should be processed more carefully as usual. That's why I decided not to rely on some storage info but do verification explicitly in every case the pallet is used. Maybe it has some overhead. But it was in my mind.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
dmitrylavrenov: Thanks, improved:
- https://github.com/humanode-network/humanode/commit/55ad986186808aad0dc49816ce9b07638f84329e
- https://github.com/humanode-network/humanode/commit/421ef86442344cb25ff3c54d8fd9898e50afad1f

crates/pallet-bridges-initializer-currency-swap/src/upgrade_init.rs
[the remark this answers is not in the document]
MOZGIII: Ok, let's discuss on a call

crates/pallet-bridges-initializer-currency-swap/src/mock/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub(crate) type EvmAccountId = u64;
+
+type Balance = u64;
+
+type BalancesInstanceNative = pallet_balances::Instance1;
+type BalancesInstanceEvm = pallet_balances::Instance2;
+
+parameter_types! {
+    pub NativeTreasury: AccountId = 4200;
+}
```
MOZGIII, line 23: Use ConstU64 instead?

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/28836f8115145581c0595c9fbc497ee92b326807

crates/pallet-bridges-initializer-currency-swap/src/mock/mod.rs
dmitrylavrenov: I prefer use `parameter_types` to indicate that it's `AccountId`, imo.

MOZGIII: My idea was to avoid sharing this bit, so we can inline it, but this works too

crates/pallet-bridges-initializer-currency-swap/src/upgrade_init.rs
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/b2d8d28e10617740d1f9e40306342c6aeb5882c6

crates/pallet-bridges-initializer-currency-swap/src/mock/v1.rs
```diff
@@ -0,0 +1,135 @@
+//! The v2 mock that includes bridges initialization logic at runtime.
```
MOZGIII, line 1:
```suggestion
//! The v1 mock that includes bridges initialization logic at runtime.
```

crates/pallet-bridges-initializer-currency-swap/src/mock/v0.rs
```diff
@@ -0,0 +1,87 @@
+//! The v1 mock that represents just a case with two separate accounts system
```
MOZGIII, line 1:
```suggestion
//! The v0 mock that represents just a case with two separate accounts system
```

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,382 @@
...
+                    + ALICE.balance
+                    + BOB.balance
+                    + EXISTENTIAL_DEPOSIT_EVM
+            );
+        });
+    })
+}
+
+#[test]
+fn initialization_idempotence() {
```
MOZGIII, line 157:
```suggestion
fn initialization_idempotency() {
```

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,382 @@
...
+                treasury.balance - (LION.balance + DOG.balance + CAT.balance + FISH.balance)
+            );
+            assert_eq!(
+                v1::EvmBalances::total_balance(&v1::SwapBridgeEvmToNativePot::account_id(),),
+                v1::Balances::total_balance(&NativeTreasury::get())
+                    + ALICE.balance
+                    + BOB.balance
+                    + EXISTENTIAL_DEPOSIT_EVM
+            );
+
```
MOZGIII, line 225:
```suggestion
assert_storage_noop!(v1::EvmNativeBridgesInitializer::initialize().unwrap());
assert_storage_noop!(v1::EvmNativeBridgesInitializer::initialize().unwrap());

```

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,382 @@
...
+            for _ in 0..5 {
+                v1::Balances::transfer(Some(ALICE.account).into(), BOB.account, 1).unwrap();
+
+                // Initialize bridges one more time.
```
MOZGIII, line 229:
```suggestion
                // Initialize bridges one more time after a change.
```

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
```diff
@@ -0,0 +1,382 @@
...
+            assert_eq!(
+                v1::EvmBalances::total_balance(&v1::SwapBridgeEvmToNativePot::account_id(),),
+                v1::Balances::total_balance(&NativeTreasury::get())
+                    + ALICE.balance
+                    + BOB.balance
+                    + EXISTENTIAL_DEPOSIT_EVM
+            );
+
+            for _ in 0..5 {
+                v1::Balances::transfer(Some(ALICE.account).into(), BOB.account, 1).unwrap();
```
MOZGIII, line 227: This does not create a new account though, but it would be a worthwhile case to check

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/55a9f8339c679b0334720a67bb966f28a1329c7e

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+                    &T::NativeEvmBridgePot::get(),
+                    current_native_bridge_balance
+                        .checked_sub(&amount)
+                        .expect("current_native_bridge_balance is greater than amount; qed."),
+                    frame_support::traits::WithdrawReasons::TRANSFER,
+                    frame_support::traits::ExistenceRequirement::KeepAlive,
+                )?;
+                weight += T::DbWeight::get().writes(1);
+
+                // We can safely ignore the result as overflow cann't be reached.
```
MOZGIII, line 280:
```suggestion
                // We can safely ignore the result as overflow can't be reached.
```

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+                    &T::NativeTreasuryPot::get(),
+                    amount
+                        .checked_sub(&current_native_bridge_balance)
+                        .expect("current_native_bridge_balance is less than amount; qed."),
+                    frame_support::traits::WithdrawReasons::TRANSFER,
+                    frame_support::traits::ExistenceRequirement::KeepAlive,
+                )?;
+                weight += T::DbWeight::get().writes(1);
+
+                // We can safely ignore the result as overflow cann't be reached.
```
MOZGIII, line 261:
```suggestion
                // We can safely ignore the result as overflow can't be reached.
```

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+            }
+            Ordering::Equal => {}
+        }
+
+        Ok(weight)
+    }
+
+    /// Make evm bridge balance be provided amount value.
+    ///
+    /// The logic shouldn't change evm swappable balance value.
```
MOZGIII, line 297:
```suggestion
    /// Make evm bridge balance be provided amount value.
    ///
    /// This function MINTS/BURNS the tokens as it needs to balance out the currencies and bridge pots.
    ///
    /// The logic shouldn't change evm swappable balance value.
```

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+
+            Ok(())
+        })?;
+
+        Ok(weight)
+    }
+
+    /// Make native bridge balance be provided amount value.
+    ///
+    /// The logic can change native swappable balance value.
```
MOZGIII, line 226:
```suggestion
    /// Make native bridge balance be provided amount value.
    ///
    /// This function TRANSFERS the tokens to/from the treasury to balance the bridge pots.
    /// It will not change the total issuance, but it can change the native swappable balance value.
```

MOZGIII: How about we add `debug_assert!`/test to this fn to make sure we don't alter the total issuance?

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
MOZGIII: How about we add `debug_assert!`/test to this fn to make sure we don't alter the swappable balance?

MOZGIII (approved):
This looks great! I think we are ready to merge, except for a couple of nits in the still unresolved comments.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
MOZGIII, line 1: Do we want to add an assertion the the resulting EVM total issuance is a nice roung number metching the native total issuance? It seems like it might not actually be the case in our implementation, but it is what we want on a high level.

MOZGIII: If the assertion fails, and we don't get the round number - I suppose the most probable reason would be going under (not over) the native balance. We can rectify this by minting more tokens and placing them into a special account whose purpose would be just that - being a "dead weight" for balancing the currencies. It should probably be another pot account.

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
dmitrylavrenov: Thanks, agree - https://github.com/humanode-network/humanode/pull/716/commits/7bacc72c823577f6d100185ad4c0433f25d64edf.

crates/pallet-bridges-initializer-currency-swap/src/lib.rs
dmitrylavrenov: Thanks, good point - https://github.com/humanode-network/humanode/pull/716/commits/d10ea31094286f52c074feea4a12887b13b70629.

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
MOZGIII: Well, yes, obviously the total issuances must be equal taking the exchange rate into account (so, an exact match for 1:1, and something else for other than 1:1 - but still proportional exactly).
It is very important to actually make it so the balances are equal/proportional with exact ratios so we don't have unnecessary irregularities - and thus it will make it so the users who don't know all the intricacies of why the total issuances have to be different sometimes don't needlessly worry.

MOZGIII: We should avoid using the storage for this, as all the ed operations are with runtime constants. Recomputing would be cheaper than invoking the storage, so *unless there is a concern on correctness that is somehow related to caching the value used for the calculation for later* - we better switch to a helper fn that doesn't read/write to storage.

crates/pallet-bridges-initializer-currency-swap/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "pallet-bridges-initializer-currency-swap"
```
MOZGIII, line 2:
```suggestion
name = "pallet-balanced-currency-swap-bridges-initializer"
```

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/43fc65c09be5a9aec23cd9b8d8f1b6e438eabd0a

crates/pallet-bridges-initializer-currency-swap/src/tests.rs
dmitrylavrenov: Agree - https://github.com/humanode-network/humanode/pull/716/commits/5075aea5262f0af79ff9f7578f8f65b284c41ed4.

crates/pallet-balanced-currency-swap-bridges-initializer/src/mock/mod.rs
```diff
@@ -0,0 +1,90 @@
+//! The mocks for the pallet testing.
+
+use frame_support::{parameter_types, sp_io, sp_runtime::BuildStorage};
+
+use crate::{self as pallet_bridges_initializer_currency_swap};
```
MOZGIII, line 5: This should be adjusted too

utils/checks/snapshots/features.yaml
```diff
@@ -1990,6 +1990,10 @@
   features:
     - default
     - std
+- name: pallet-bridges-initializer-currency-swap 0.1.0
```
MOZGIII, line 1993: And this

crates/pallet-balanced-currency-swap-bridges-initializer/src/mock/mod.rs
dmitrylavrenov: My bad, thanks - https://github.com/humanode-network/humanode/pull/716/commits/a2263507d50bf848858c70768f4432e2d24fe3ec.

utils/checks/snapshots/features.yaml
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/682cb95d89c03e73be78520429e1b72e7eef9f5e

crates/pallet-balanced-currency-swap-bridges-initializer/src/tests.rs
MOZGIII, line 1: See the discussing here: https://github.com/humanode-network/humanode/commit/a5876adbb4c2b34288fe34daf3647ae725519df1#r124524318

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/716/commits/7c7f03108e41dbf70df33954e09657f852c8567c

MOZGIII (approved):
Great! Last minute finding: now the only thing lacking is proper test descriptions - for what each test aims to verify. Let's add that and we can merge!

dmitrylavrenov:
> Great! Last minute finding: now the only thing lacking is proper test descriptions - for what each test aims to verify. Let's add that and we can merge!

Sure, check please - https://github.com/humanode-network/humanode/pull/716/commits/ea09ee7f72cf5fb2cbc4061344854d1362d1ca42.

PR #746, opened 2023-08-24
Add eHMND ERC20 precompile

### LINKED ISSUE #696: Add ERC20 wrapper precompile
We need something like this: https://github.com/PureStake/moonbeam/blob/master/precompiles/balances-erc20/ERC20.sol

dmitrylavrenov:
Closes #696


crates/precompile-evm-balances-erc20/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+            Action::Transfer => Self::transfer(handle),
+            Action::TransferFrom => Self::transfer_from(handle),
+        }
+    }
+}
+
+impl<EvmBalancesT, Metadata, GasCost> EvmBalancesErc20<EvmBalancesT, Metadata, GasCost>
+where
+    Metadata: Erc20Metadata,
+    EvmBalancesT: pallet_evm_balances::Config,
```
MOZGIII, line 137: Can't we use a generic `Currency` here?

MOZGIII (commented):
Let's add an E2E test now that it is easy!

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -71,3 +80,19 @@ where
...
+
+pub struct EvmBalancesErc20Metadata;
+
+impl precompile_evm_balances_erc20::Erc20Metadata for EvmBalancesErc20Metadata {
+    fn name() -> &'static str {
+        "Ethereum humanode token"
+    }
+
+    fn symbol() -> &'static str {
+        "eHMND"
```
MOZGIII, line 92:
```suggestion
        "WeHMND"
```

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -71,3 +80,19 @@ where
 fn hash(a: u64) -> H160 {
     H160::from_low_u64_be(a)
 }
+
+pub struct EvmBalancesErc20Metadata;
+
+impl precompile_evm_balances_erc20::Erc20Metadata for EvmBalancesErc20Metadata {
+    fn name() -> &'static str {
+        "Ethereum humanode token"
```
MOZGIII, line 88:
```suggestion
        "Wrapped eHMND"
```

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -71,3 +80,19 @@ where
...
+    }
+
+    fn symbol() -> &'static str {
+        "eHMND"
+    }
+
+    fn decimals() -> u8 {
+        18
```
MOZGIII, line 96: Instead of hardcoding this here, better refer to the currency

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -51,6 +55,11 @@ where
             // Humanode precompiles:
             a if a == hash(2048) => Some(Bioauth::<R>::execute(handle)),
             a if a == hash(2049) => Some(EvmAccountsMapping::<R>::execute(handle)),
+            a if a == hash(2050) => {
```
MOZGIII, line 58: Why `2050`?

crates/precompile-evm-balances-erc20/ERC20.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: UNLICENSED
+
+pragma solidity >=0.7.0 <0.9.0;
+
+/**
+ * @title ERC20 Interface
+ *
+ * An interface exposing native eHMND tokens as ERC20 tokens.
+ *
+ * Address: 0x0000000000000000000000000000000000000802
```
MOZGIII, line 10: Does this value match the one here https://github.com/humanode-network/humanode/pull/746/files#r1306311603?

crates/precompile-evm-balances-erc20/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+    const STORAGE_PREFIX: &'static str = "Approves";
+
+    fn pallet_prefix() -> &'static str {
+        "EvmBalancesErc20Approves"
+    }
+}
+
+/// Storage type used to store approvals, since `pallet_evm_balances` doesn't handle this behavior.
+/// (Owner => Allowed => Amount).
+pub type ApprovesStorage<Runtime> = StorageDoubleMap<
```
MOZGIII, line 58: Can we have a pallet for this instead? It will sure come in handy for a more advanced management of the state, like upgrades and etc, if we need it later.

We can call it `pallet_erc20_approvals` or `pallet_erc20`

MOZGIII (commented):
Looks good overall!

I have some thoughts regarding the general code organization. I think it would make sense to add a pallet in addition to the precompile. The precompile can then rely on the pallet-provided structures. So, the precompile would essentially have a dependency on the pallet.

Also, it seems like we could very much just use the `Currency` interface instead of having a strict dependency on the evm balances pallet.

MOZGIII:
Another note is we probably want to apply some guardrails to improve the UX a bit. We should reject the transfers of the ERC20 tokens to the address of the precompile itself. The rationale here is it is meaningless to send tokens in such a way (they'll be lost) but there is a non-zero chance that a user will use the contract address where the destination address is supposed to go. We can easily prevent this from ever happening if we add a check.

MOZGIII:
Another note: I suppose we will be implementing other standard on the same precompile address in the future (i.e. ERC777 in  addition to the ERC20). It may be wise to name things in the manner of `precompile-wrapped-currency` instead of `precompile-erc20` to be future-proof and allow ourselves flexibility later on.

crates/humanode-runtime/src/frontier_precompiles.rs
dmitrylavrenov: `EvmAccountsMapping` has 2049.  2050 is `+1` as next evm related precompile.

MOZGIII: Can/should we use the same value at they use at moonbeam?

crates/pallet-erc20/src/lib.rs
MOZGIII, line 1: Makes sense to move not just the approvals storage here, but also the approvals handling logic, and well as transfers.

crates/precompile-evm-balances-erc20/src/lib.rs
MOZGIII, line 1: Ideally, we should move the logic to the pallet, and only use precompiles to implement the solidity call convention while delegating to the pallet for the actual implementation.

crates/humanode-runtime/src/frontier_precompiles.rs
dmitrylavrenov: It's the same value as well :)

crates/precompile-evm-balances-erc20/src/lib.rs
dmitrylavrenov: Sure, good point, moved to `pallet-token-wrapper` and `Currency` is used as a part of `Config`.

crates/precompile-evm-balances-erc20/src/lib.rs
dmitrylavrenov: Sure, `pallet-token-wrapper` has been added.

crates/precompile-evm-balances-erc20/src/lib.rs
dmitrylavrenov: Thanks, the logic has been moved.

crates/humanode-runtime/src/lib.rs
```diff
@@ -832,6 +854,7 @@ construct_runtime!(
         EvmToNativeSwapBridgePot: pallet_pot::<Instance5> = 34,
         CurrencySwap: pallet_currency_swap = 35,
         BalancedCurrencySwapBridgesInitializer: pallet_balanced_currency_swap_bridges_initializer = 36,
+        EvmBalancesErc20: pallet_wrapped_token = 37,
```
MOZGIII, line 857:
```suggestion
        WrappedEvmBalances: pallet_wrapped_token = 37,
```

crates/pallet-wrapped-token/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Returns the amount of tokens owned by provided account.
+    pub fn balance_of(owner: &AccountIdOf<T, I>) -> BalanceOf<T, I> {
+        T::Currency::total_balance(owner)
+    }
+
+    /// Returns the remaining number of tokens that spender will be allowed to spend on behalf of
+    /// owner. This is zero by default.
+    pub fn allowance(owner: &AccountIdOf<T, I>, spender: &AccountIdOf<T, I>) -> BalanceOf<T, I> {
```
MOZGIII, line 107: I think there are aliases that can do just this, can't they? They might be deprecated in the substrate `master` though, so not sure if we want to use them... But just to note this. Maybe double-check if they are indeed deprecated, and then use them if they are not?

crates/pallet-wrapped-token/src/lib.rs
```diff
@@ -0,0 +1,165 @@
+//! A substrate pallet that wraps funds related instance.
```
MOZGIII, line 1: This comment does not provide a sufficient description for this pallet.

crates/precompile-wrapped-token/src/lib.rs
```diff
@@ -0,0 +1,339 @@
+//! A precompile to interact with wrapped funds related instances.
```
MOZGIII, line 1: It is not clear what this does.

MOZGIII (commented):
This looks great! I think I'll have to take another look, but I already left some comments.

What do you think about moving the implementation (i.e. the pallet contents) to pallet evm balances? Not saying that we should do it, but maybe we should think this through.

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/746/commits/9275197b7c465b5943d5b6878b372fc9fb0ce0ed

crates/precompile-utils/src/testing.rs
MOZGIII, line 1: How about doing this in a separate PR?

dmitrylavrenov: Moved - https://github.com/humanode-network/humanode/pull/767

crates/pallet-wrapped-token/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/746/commits/654dc6523c41bb6598fa23e34b3ca80dcec31b42

crates/pallet-wrapped-token/src/lib.rs
dmitrylavrenov: Edited.

crates/precompile-wrapped-token/src/lib.rs
dmitrylavrenov: Edited.

crates/humanode-runtime/src/lib.rs
```diff
@@ -788,6 +788,28 @@ impl pallet_balanced_currency_swap_bridges_initializer::Config for Runtime {
     type WeightInfo = ();
 }
 
+pub struct Erc20EvmBalancesMetadata;
```
MOZGIII, line 791:
```suggestion
pub struct EvmBalancesErc20Metadata;
```

crates/humanode-runtime/src/lib.rs
```diff
@@ -832,6 +854,7 @@ construct_runtime!(
         EvmToNativeSwapBridgePot: pallet_pot::<Instance5> = 34,
         CurrencySwap: pallet_currency_swap = 35,
         BalancedCurrencySwapBridgesInitializer: pallet_balanced_currency_swap_bridges_initializer = 36,
+        Erc20EvmBalances: pallet_erc20_support = 37,
```
MOZGIII, line 857:
```suggestion
        EvmBalancesErc20Support: pallet_erc20_support = 37,
```

crates/humanode-runtime/src/lib.rs
```diff
@@ -788,6 +788,28 @@ impl pallet_balanced_currency_swap_bridges_initializer::Config for Runtime {
...
+    fn symbol() -> &'static str {
+        "WeHMND"
+    }
+
+    fn decimals() -> u8 {
+        18
+    }
+}
+
+impl pallet_erc20_support::Config for Runtime {
```
MOZGIII, line 807: This should be instanced since it could be used with multiple currencies in theory

MOZGIII (commented):
Can we also add E2E tests for this? Could be in another PR...

MOZGIII (commented):
Can we also take https://eips.ethereum.org/EIPS/eip-6093 into consideration?

MOZGIII (approved):
Looks good, I'll take another look at tests next, but I think I can start shipping this to the internal test.

crates/precompile-erc20-support/Cargo.toml
MOZGIII, line 1: Precompile has to encompass not just ERC20, but all the future ERCs we'll add here too. It's not like we can add more than one (i.e. implement ERC777 in another precompile - it has to be this precompile to work).

Support pallets can be multiple.

To sum up, rename the precompile to `precompile-native-token` or `precompile-native-currency` (but leave the `pallet-erc20-support` name as-is).

utils/checks/snapshots/features.yaml
```diff
@@ -247,6 +247,11 @@
     - default
     - sha2
     - std
+- name: bstr 0.2.17
```
MOZGIII, line 250: Where does this come from?

crates/pallet-wrapped-token/src/lib.rs
MOZGIII: This is not what I meant...

utils/checks/snapshots/features.yaml
dmitrylavrenov: If I remember correctly, it has been added after using `testing` feature at precompile code.
image

crates/pallet-erc20-support/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Returns the remaining number of tokens that spender will be allowed to spend on behalf of
+    /// owner. This is zero by default.
+    pub fn allowance(owner: &AccountIdOf<T, I>, spender: &AccountIdOf<T, I>) -> BalanceOf<T, I> {
+        Self::approvals(owner, spender).unwrap_or_default()
+    }
+
+    /// Sets amount as the allowance of spender over the caller’s tokens.
+    pub fn approve(owner: AccountIdOf<T, I>, spender: AccountIdOf<T, I>, amount: BalanceOf<T, I>) {
+        <Approvals<T, I>>::insert(owner, spender, amount);
```
MOZGIII, line 114: Should we prohibit approvals where `owner == spender`?

crates/precompile-erc20-support/Cargo.toml
dmitrylavrenov: Got your thoughts - https://github.com/humanode-network/humanode/pull/746/commits/d66cddff5b75573110f6c77a1a01ceee803b0a6f.

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/746/commits/78b3097b6b44e6ce3fdf05f73376cfa6dad3f934

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/746/commits/974d18803b800f97425e3f29ac9ea6629c8d3b72

crates/pallet-erc20-support/src/lib.rs
dmitrylavrenov: I guess we can prohibit it but in general if we check openzeppelin ERC20 implementation - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/618304cc01e718cbdd87059ce2978b748c15050f/contracts/token/ERC20/ERC20.sol#L283. This case is not prohibited.

dmitrylavrenov:
> Can we also add E2E tests for this? Could be in another PR...

Will add in separate PR. The issue has been created - https://github.com/humanode-network/humanode/issues/775.

dmitrylavrenov:
> Can we also take https://eips.ethereum.org/EIPS/eip-6093 into consideration? Also, could be in another PR.

Let's add in separate PR to simplify current PR. The issue has been created - https://github.com/humanode-network/humanode/issues/776.

crates/pallet-wrapped-token/src/lib.rs
dmitrylavrenov: Reverted - https://github.com/humanode-network/humanode/pull/746/commits/66f9c506b78259c18924508587679968da313511.

crates/pallet-erc20-support/src/lib.rs
MOZGIII: But then they also don't have the aspect that the owner doesn't consume allowance, like we have: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/618304cc01e718cbdd87059ce2978b748c15050f/contracts/token/ERC20/ERC20.sol#L311

There is owner uses `transferFrom` but doesn't create an allowance for themselves they'll get a rever just like anyone else would.
Out implementation, on the other hand, has the `If caller is "sender", it can spend as much as it wants.` commend and the associated logic, which doesn't seem right. I think we should rather make it go through the usual allowance restrictions, even for the owner.

dmitrylavrenov: Got it, added the restriction logic for the same owner and spender - https://github.com/humanode-network/humanode/pull/746/commits/78e8bf8fb04bd40300a160ac3d84679c57ca21fa.

MOZGIII: You have to pick one, not both. :D

Either we make allowances limit the spending for the owner (via the `transferFrom` calls only!) - in which case the owner can add them for themselves, or we do not allow the allowances to limit owners' spending when using `transferFrom` - then we must not allow adding the allowances for the owner in the first place. I didn't look into the EIPs or anything, but this is what just makes sense logically. I suggest checking the ERC20 spec - maybe it explicitly specifies what has to happen, so we don't have to choose/guess here.

MOZGIII (commented):
This looks good so far! I'll take some more time for the final review today

MOZGIII (approved):

PR #777, opened 2023-09-18
Dummy contract code for EVM precompiles

dmitrylavrenov:


crates/pallet-evm-precompiles-constant-accounts/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "pallet-evm-precompiles-constant-accounts"
```
MOZGIII, line 2:
```suggestion
name = "pallet-dummy-precompiles-code"
```

crates/pallet-evm-precompiles-constant-accounts/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Configuration trait of this pallet.
+    #[pallet::config]
+    pub trait Config: frame_system::Config + pallet_evm::Config {}
+
+    #[pallet::genesis_config]
+    #[derive(Default)]
+    pub struct GenesisConfig {
+        /// The list of precompiles to be stored as account codes at evm.
+        pub precompiles: Vec<(H160, Vec<u8>)>,
+    }
```
MOZGIII, line 41:
```suggestion
    pub struct GenesisConfig {
    }
```

We should instead configure the set of accounts via the `Config` (via a `Get<Vec<H160>>`).

There also should be just a built-in code in this pallet uses, no need to make it configurable.

crates/pallet-evm-precompiles-constant-accounts/Cargo.toml
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/aca209215bbac45a54129807adfee70e214fd375

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+mod tests;
+
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+/// The current precompiles addresses creation version.
+pub const CURRENT_CREATION_VERSION: u16 = 1;
+
+/// The dummy code used to be stored for precompiles addresses.
+pub const DUMMY_CODE: &str = "DUMMY_CODE";
```
MOZGIII, line 22: Instead of an invalid opcodes sequence, I suggest using the following value: `0x5F5FFD` (as raw bytes of course).

So, what this is, and how did I came up with it? First: why. Just is case someone or something tries to interpert the code to see if it is valid.

This is actually a hand-crafted sequence of opcodes for a bare-bones revert. What I did was I reasoned that it would make sense for this dummy code to actually fail the execution somehow, and I chose revert instead of error as a less harmful option.
Then, I just looked up the `REVERT` opcode (which is `FD`) - it takes two arguments from the stack, so I push two zeroes to the stack with `PUSH0` (`5F` twice). That's pretty much it.

crates/humanode-runtime/src/lib.rs
```diff
@@ -810,6 +810,14 @@ impl pallet_erc20_support::Config for Runtime {
     type Metadata = EvmBalancesErc20Metadata;
 }
 
+frame_support::parameter_types! {
+    pub PrecompilesAddresses: Vec<H160> = vec![frontier_precompiles::hash(2050)];
```
MOZGIII, line 814: Do we only want `2050` (which is what? I suggest we declare a more meaningful way of referencing precompiles, for instance a `mod` with a bunch of `pub const QWE: ... = 2050;`-ish items at the `frontier_precompiles.rs`).

MOZGIII (approved):
Looks good! I left some comments with proposed improvements, once those are addressed we can merge I think

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: I was thinking that `2050` is the only now to stored as code. Should we add more?

Precompiles constants have been added to simplify references - https://github.com/humanode-network/humanode/pull/777/commits/42d81213e82ccbd63220788e604d6a1851fac7b4.

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: Great suggestion, really like it - https://github.com/humanode-network/humanode/pull/777/commits/52251057df374e623c9c75e67ccc36c5cfce63fa.

crates/humanode-runtime/src/lib.rs
MOZGIII: I suppose the standard precompiles don't need this, but anything that pretends to be a contract (emulates contract ABI and calls) and does not have a Solidity built-in syntax sugar (like `ecrecover`, et al) the dummy code would be helpful.

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+mod tests;
+
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+/// The dummy code used to be stored for precompiles addresses - 0x5F5FFD (as raw bytes).
+///
+/// This is actually a hand-crafted sequence of opcodes for a bare-bones revert.
+/// The REVERT opcode (which is FD) - it takes two arguments from the stack with PUSH0 (5F twice).
+pub const DUMMY_CODE: [u8; 3] = [95, 95, 253];
```
MOZGIII, line 22: We should use the hex encoding here at the very least:

```suggestion
pub const DUMMY_CODE: [u8; 3] = [0x5F, 0x5F, 0xFD];
```

Maybe we could even use `hex_literal` macro here - would be even better.

Also, can't we do `[u8; _]` instead of `[u8; 3]`?

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        fn on_runtime_upgrade() -> Weight {
+            let creation_version = Self::creation_version();
+            let mut weight = T::DbWeight::get().reads(1);
+
+            if creation_version != T::CreationVersion::get() {
+                for precompile_address in &T::PrecompilesAddresses::get() {
+                    let code = pallet_evm::Pallet::<T>::account_codes(*precompile_address);
+                    weight += T::DbWeight::get().reads(1);
+
+                    if code != DUMMY_CODE.to_vec() {
```
MOZGIII, line 78:
```suggestion
                    if &code != &DUMMY_CODE {
```

Try avoiding needless `Vec`s to save on `malloc`s.

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/ec1e1b0528b6ee3263f783e075ca6c40a7c6835c

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/7c8287454bada406a0c9902608a9746762ea19ad

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/f0184dd0a7f4c37c675624de581f92f171486a5b

crates/humanode-runtime/src/lib.rs
```diff
@@ -810,6 +810,15 @@ impl pallet_erc20_support::Config for Runtime {
     type Metadata = EvmBalancesErc20Metadata;
 }
 
+frame_support::parameter_types! {
+    pub PrecompilesAddresses: Vec<H160> = vec![frontier_precompiles::hash(precompiles_constants::NATIVE_CURRENCY)];
+}
+
+impl pallet_dummy_precompiles_code::Config for Runtime {
+    type CreationVersion = ConstU16<1>;
```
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/a2a03d705bd9c0ac1a9db13e1c4c4eb0cc597973

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub struct Pallet<T>(_);
+
+    /// Configuration trait of this pallet.
+    #[pallet::config]
+    pub trait Config: frame_system::Config + pallet_evm::Config {
+        /// The list of precompiles adresses to be created at evm with dummy code.
+        type PrecompilesAddresses: Get<Vec<H160>>;
+
+        /// The last force update ask counter.
+        type LastForceUpdateAskCounter: Get<u16>;
```
MOZGIII, line 52:
```suggestion
        type ForceUpdateAskCounter: Get<u16>;
```

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// The creation version.
+    #[pallet::storage]
+    #[pallet::getter(fn creation_version)]
+    pub type CreationVersion<T: Config> = StorageValue<_, u16, ValueQuery>;
```
MOZGIII, line 58:
```suggestion
    pub type LastCreationVersion<T: Config> = StorageValue<_, u16, ValueQuery>;
```

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/576de9ec8526a2eecd18891df342f19b90d0de19

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+                weight += T::DbWeight::get().writes(1);
+            } else if last_force_update_ask_counter != T::LastForceUpdateAskCounter::get() {
+                weight += Self::precompiles_addresses_add_dummy_code();
+
+                <LastForceUpdateAskCounter<T>>::put(T::LastForceUpdateAskCounter::get());
+                weight += T::DbWeight::get().writes(1);
+            }
+
+            weight
+        }
```
MOZGIII:
```suggestion
        fn on_runtime_upgrade() -> Weight {
            let current_force_update_ask_counter = T::ForceUpdateAskCounter::get();
            
            let last_creation_version = Self::last_creation_version();
            let last_force_update_ask_counter = Self::last_force_update_ask_counter();
            let mut weight = T::DbWeight::get().reads(2);

            let is_version_mismatch = last_creation_version != CURRENT_CREATION_VERSION;
            let is_forced = last_force_update_ask_counter != current_force_update_ask_counter;
            
            if is_version_mismatch || is_forced {
                weight += Self::precompiles_addresses_add_dummy_code();

                <CreationVersion<T>>::put(CURRENT_CREATION_VERSION);
                <LastForceUpdateAskCounter<T>>::put(T::LastForceUpdateAskCounter::get());
                weight += T::DbWeight::get().writes(1);
            }

            weight
        }
```

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+/// The current precompiles addresses creation version.
+pub const CURRENT_CREATION_VERSION: u16 = 1;
```
MOZGIII, line 22: On the semantics of the "creation" here and elsewhere. It doesn't seem flexible enough, as we might want to not just create the accounts, but also update them later.
I suggest `CURRENT_PROCESSING_VERSION`, `CURRENT_EXECUTION_VERSION`, `CURRENT_RUN_VERSION` or something like that.

Obviously, if we change this here we'll have to sync everything else accordingly, i.e. `LastExecutionVersion`, local vars names in the code, etc.

We might also apply a similar change to the `ForceUpdateAskCounter`: `ForceExecuteAskCounter` and so on.

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// A helper function to add dummy code for provided precompiles adrresses.
+    fn precompiles_addresses_add_dummy_code() -> Weight {
+        let mut weight = T::DbWeight::get().reads(0);
+
+        for precompile_address in &T::PrecompilesAddresses::get() {
+            let code = pallet_evm::Pallet::<T>::account_codes(*precompile_address);
+            weight += T::DbWeight::get().reads(1);
+
+            if code != DUMMY_CODE {
+                pallet_evm::Pallet::<T>::create_account(*precompile_address, DUMMY_CODE.to_vec());
```
MOZGIII, line 142: Don't we need a system account here too?

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/fbbdb153c4497395bdc4df08c15c8fea2f65a626

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/fe4badfcd2400e1e87710ff316c237f2bdee821b

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/d76f271132fc1a26bb5900af110d5d04cc584de4

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/777/commits/bd10f82205bfbf7652d327c60c84e8866ec45eaa

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: Which exactly system account are you talking about?

MOZGIII: `frame_system` / `evm_system` - depending on what's used in `pallet_evm`

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> Weight {
+            let last_execution_version = Self::last_execution_version();
+            let last_force_execute_ask_counter = Self::last_force_execute_ask_counter();
+
+            let current_force_execute_ask_counter = T::ForceExecuteAskCounter::get();
+            let mut weight = T::DbWeight::get().reads(3);
+
+            let is_version_mismatch = last_execution_version != CURRENT_EXECUTION_VERSION;
+            let is_forced = last_force_execute_ask_counter != current_force_execute_ask_counter;
```
MOZGIII, line 91:
```suggestion
            let is_forced = last_force_execute_ask_counter < current_force_execute_ask_counter;
```

crates/pallet-dummy-precompiles-code/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> Weight {
+            let last_execution_version = Self::last_execution_version();
+            let last_force_execute_ask_counter = Self::last_force_execute_ask_counter();
+
+            let current_force_execute_ask_counter = T::ForceExecuteAskCounter::get();
+            let mut weight = T::DbWeight::get().reads(3);
```
MOZGIII, line 88:
```suggestion
            let mut weight = T::DbWeight::get().reads(2);
```

Only two reads from the DB.

MOZGIII (approved):
Left a few comments, other than that looks good

crates/pallet-dummy-precompiles-code/src/lib.rs
dmitrylavrenov: It's managed under `create_account` fn - https://github.com/humanode-network/frontier/blob/8d3a2ef77867d1e40bd42078eb371436d49f0b1d/frame/evm/src/lib.rs#L705.

PR #780, opened 2023-09-18
Set EVM currency existential deposit to 1

### LINKED ISSUE #781: Set EVM currency existenial deposit to 1

MOZGIII:
Closes #781.

- #781

Waits for #785 to land before merging.

### DIFF
--- a/crates/humanode-runtime/src/lib.rs
+++ b/crates/humanode-runtime/src/lib.rs
@@ -658,7 +658,7 @@ impl pallet_evm_balances::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
     type AccountId = EvmAccountId;
     type Balance = Balance;
-    type ExistentialDeposit = ConstU128<500>;
+    type ExistentialDeposit = ConstU128<1>;
     type AccountStore = EvmSystem;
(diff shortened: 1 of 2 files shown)


dmitrylavrenov:
Should we research and properly run migrations at this PR?

MOZGIII:
> Should we research and properly run migrations at this PR?

Let's discuss

dmitrylavrenov:
Spent time on research in a little:
- `Balances` and `EvmBalances` don't have some migration logic. Anyway, in case new ED is less than an older one then it doesn't influence on accounts creation/deletion during runtime upgrade
- Checked that we should update bridges initializer version - https://github.com/humanode-network/humanode/commit/53f319e7680bc065ee4dfbf924e6092a6211477f

MOZGIII:
So, this should just work, right? We'll have to test extensively through try-runtime and on the internal testing stages before we ship even to testnet.

dmitrylavrenov:
> So, this should just work, right? I don't see why not... However, we'll have to test extensively through try-runtime and on the internal testing stages before we ship even to testnet.

Yes, this should. Right, agree with the way of testing.

dmitrylavrenov (approved):

PR #785, opened 2023-09-20
Allow force-rebalancing bridges

dmitrylavrenov:


crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
```diff
@@ -39,9 +39,6 @@ type EvmBalanceOf<T> =
 /// The current storage version.
 const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
 
-/// The current bridges initializer version.
-pub const CURRENT_BRIDGES_INITIALIZER_VERSION: u16 = 1;
```
MOZGIII, line 43: I think we should still keep an internal version though. It just wouldn't be bumped this time, but the reason why we introduced it in the first place - to keep track of which algorithm we used to balance out the bridges - it still very valid.

crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
```diff
@@ -96,6 +93,13 @@ pub mod pallet {
         /// The evm-native bridge pot account.
         type EvmNativeBridgePot: Get<Self::EvmAccountId>;
 
+        /// The current bridges initializer version.
+        type InitializerVersion: Get<u16>;
```
MOZGIII, line 97: To avoid the confusion with the `CURRENT_BRIDGES_INITIALIZER_VERSION` and the version of the pallet state semantics we should call this something else.

crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
```diff
@@ -96,6 +93,13 @@ pub mod pallet {
...
+
+        /// Whether currencies balanced check required on runtime upgrade.
+        #[pallet::constant]
+        type IsBalancedCheckRequiredOnRuntimeUpgrade: Get<bool>;
```
MOZGIII, line 101: This should be more expressive. What happens if the check is not required? Is it skipped? Let's just call it `SkipBalanceVerificationOnRuntimeUpgrade`

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
```diff
@@ -22,7 +22,14 @@ pub fn on_runtime_upgrade<T: Config>() -> Weight {
             }
         }
 
-        <InitializerVersion<T>>::put(CURRENT_BRIDGES_INITIALIZER_VERSION);
+        <InitializerVersion<T>>::put(T::InitializerVersion::get());
+    } else if T::IsBalancedCheckRequiredOnRuntimeUpgrade::get() {
```
MOZGIII, line 26: Doesn't look like this is the logic we have discussed, so we'll need to discuss this again.

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
```diff
@@ -4,14 +4,14 @@ use frame_support::pallet_prelude::*;
...
-use crate::{Config, InitializerVersion, Pallet, CURRENT_BRIDGES_INITIALIZER_VERSION};
+use crate::{Config, InitializerVersion, Pallet};
 
 /// Initialize the bridges pot accounts.
 pub fn on_runtime_upgrade<T: Config>() -> Weight {
     let initializer_version = <InitializerVersion<T>>::get();
     let mut weight = T::DbWeight::get().reads(1);
 
-    if initializer_version != CURRENT_BRIDGES_INITIALIZER_VERSION {
+    if initializer_version < T::InitializerVersion::get() {
```
dmitrylavrenov, line 14:
```suggestion
    if initializer_version < T::ManualRebalanceAskCounter::get() {
```

dmitrylavrenov: Renamed

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/785/commits/73c28576205966c2e867eebe0280a511778dc8f8

crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
dmitrylavrenov: Renamed.

crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
dmitrylavrenov: Redesigned based on discussion.

crates/pallet-balanced-currency-swap-bridges-initializer/src/lib.rs
dmitrylavrenov: Returned.

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
```diff
@@ -4,25 +4,31 @@ use frame_support::pallet_prelude::*;
...
-        if !is_balanced {
-            match Pallet::<T>::initialize() {
-                Ok(w) => weight += w,
-                Err(err) => sp_tracing::error!("error during bridges initialization: {err:?}"),
-            }
+    let last_initializer_version = <LastInitializerVersion<T>>::get();
+    let last_force_rebalance_ask_counter = <LastForceRebalanceAskCounter<T>>::get();
+    let current_force_rebalance_ask_counter = T::ForceRebalanceAskCounter::get();
+
+    let mut weight = T::DbWeight::get().reads(3);
```
MOZGIII, line 18:
```suggestion
    let mut weight = T::DbWeight::get().reads(2);
```

`T::ForceRebalanceAskCounter::get();` is not a db read

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
```diff
@@ -4,25 +4,31 @@ use frame_support::pallet_prelude::*;
...
+
+    let is_version_mismatch = last_initializer_version != CURRENT_BRIDGES_INITIALIZER_VERSION;
+    let is_forced = last_force_rebalance_ask_counter != current_force_rebalance_ask_counter;
```
MOZGIII, line 21:
```suggestion
    let is_forced = last_force_rebalance_ask_counter < current_force_rebalance_ask_counter;
```

MOZGIII (approved):

PR #809, opened 2023-10-19
Introduce `Allowance` type as a part of `Config` at `pallet-erc20-support`

dmitrylavrenov:


crates/pallet-erc20-support/src/lib.rs
```diff
@@ -65,6 +70,17 @@ pub mod pallet {
         /// The currency to be exposed as ERC20 token.
         type Currency: Currency<AccountIdOf<Self, I>>;
 
+        /// Allowance balance type.
+        type AllowanceBalance: From<BalanceOf<Self, I>>
+            + Sub<Output = Self::AllowanceBalance>
```
MOZGIII, line 75:
```suggestion
            + Sub
```

I think `Sub` actually has a default that would work for use here. Also, why not `CheckedSub`?

crates/pallet-erc20-support/src/lib.rs
```diff
@@ -65,6 +70,17 @@ pub mod pallet {
...
+            + PartialOrd
+            + Default
+            + Copy
+            + Codec
+            + EncodeLike
+            + MaxEncodedLen
+            + scale_info::TypeInfo;
```
MOZGIII, line 82: Some of these can be replaced with `Parameter + Member`

crates/pallet-erc20-support/src/lib.rs
dmitrylavrenov: Unfortunately, `U256` and `U512` don't implement `CheckedSub`.

crates/pallet-erc20-support/src/lib.rs
```diff
@@ -65,6 +70,17 @@ pub mod pallet {
         /// The currency to be exposed as ERC20 token.
         type Currency: Currency<AccountIdOf<Self, I>>;
 
+        /// Allowance balance type.
+        type AllowanceBalance: From<BalanceOf<Self, I>>
```
MOZGIII, line 74:
```suggestion
        type Allowance: From<BalanceOf<Self, I>>
```

The allowance does not represent a balance in the sense this is not the balance of the operations. It is more like a credit.

crates/pallet-erc20-support/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/635dfa44fc93abf585cfb822911839161d65a2d8

crates/pallet-erc20-support/src/lib.rs
```diff
@@ -142,9 +165,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
                 // Remove "value" from allowed, exit if underflow.
-                let allowed = entry
-                    .checked_sub(&amount)
-                    .ok_or(Error::<T, I>::SpendMoreThanAllowed)?;
+
+                if T::AllowanceBalance::from(amount) > *entry {
+                    return Err(Error::<T, I>::SpendMoreThanAllowed);
+                }
+
+                let allowed = entry.sub(amount.into());
```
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/4d16fe611783bfcdca5cc9e91019cab7a96ae3fd

crates/pallet-erc20-support/src/lib.rs
MOZGIII: Let's create a tech-debt issue then!

crates/pallet-erc20-support/Cargo.toml
```diff
@@ -9,12 +9,21 @@ codec = { workspace = true, package = "parity-scale-codec", features = ["derive"
 frame-support = { workspace = true }
 frame-system = { workspace = true }
 scale-info = { workspace = true, features = ["derive"] }
+sp-std = { workspace = true }
```
MOZGIII, line 12:
```suggestion
```

This should not be required since we can reach it via `frame_support::sp_std`.

crates/pallet-erc20-support/src/migrations/v1.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    // Read the onchain version.
+    let mut weight: Weight = T::DbWeight::get().reads(1);
+
+    info!(message = "Running migration to v1", from = ?onchain);
+
+    if onchain == 1 {
+        info!(message = "Already at version 1, nothing to do");
+        return weight;
+    }
```
MOZGIII, line 23: What is `onchain` is 2? I think this migration is only safe to run when the `onchain` version is `0`, not when it's not `1`.

crates/pallet-erc20-support/src/migrations/v1.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/// Check the state after the migration.
+///
+/// Panics if anything goes wrong.
+#[cfg(feature = "try-runtime")]
+pub fn post_migrate<T: Config<I>, I: 'static>(_state: Vec<u8>) {
+    // Ensure version is updated correctly.
+    let onchain = <Pallet<T, I>>::on_chain_storage_version();
+    assert_eq!(onchain, 1);
```
MOZGIII, line 52: This might not hold if we are migrating on version `2`, in which case the migration will be a noop, but the onchain version will not be `1` (but `2`).

crates/pallet-erc20-support/src/tests.rs
MOZGIII, line 1: How about adding a test so ensure we can set the `U256::MAX-1` allowance?

crates/precompile-native-currency/src/lib.rs
```diff
@@ -211,7 +213,7 @@ where
             owner.into(),
             spender.into(),
             amount.try_into().map_err(|_| PrecompileFailure::Error {
-                exit_status: ExitError::Other("value is out of bounds".into()),
+                exit_status: ExitError::Other("allowance convertation error from U256".into()),
```
MOZGIII, line 216:
```suggestion
                exit_status: ExitError::Other("allowance is out of bounds".into()),
```

MOZGIII: Actually, how about actually matching the error here, or maybe even doing a `.to_string`?

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: There is no `onchain` 2. The migration from `0` to `1`.

crates/pallet-erc20-support/Cargo.toml
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/9673046a9a74da360b77cbb0b3aed8fd438ae40e

crates/pallet-erc20-support/src/migrations/v1.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    if onchain == 1 {
+        info!(message = "Already at version 1, nothing to do");
+        return weight;
+    }
+
+    <Approvals<T, I>>::translate(|_owner, _spender, amount: BalanceOf<T, I>| Some(amount.into()));
+
+    // Set new version.
+    current.put::<Pallet<T, I>>();
```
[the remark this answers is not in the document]
dmitrylavrenov: But the same approach is used at `pallet-humanode-session`? Btw, it double checks that new version has been stored.

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: We are migration from `0` to `1`.

crates/pallet-erc20-support/src/tests.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/42f8f096f4ca81bf0970799c5d0d81d9fc6eb30e

crates/pallet-erc20-support/src/tests.rs
```diff
@@ -44,7 +45,29 @@ fn approve_works() {
...
+        assert_eq!(
+            Erc20Balances::approvals(alice, bob),
+            approved_balance.into()
+        );
+    })
+}
+
+/// This test verifies that approval logic works as expected in case approving `U256::MAX-1` value.
+#[test]
+fn approve_almost_max_works() {
```
MOZGIII, line 63:
```suggestion
fn approve_max_works() {
```

This is in fact the maximum possible approval for the `Mock` config.

crates/precompile-native-currency/src/lib.rs
dmitrylavrenov: I would suggest to use explicit `allowance is out of bounds` here as default error message for `TryFrom<U256>` is just `Overflow` without any details.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/e9d0e8cfadfb50cd099611d55745e195fcd2d266

crates/precompile-native-currency/src/lib.rs
```diff
@@ -115,6 +119,7 @@ where
     Erc20SupportT: pallet_erc20_support::Config,
     AccountIdOf<Erc20SupportT>: From<H160>,
     BalanceOf<Erc20SupportT>: Into<U256> + TryFrom<U256>,
+    AllowanceOf<Erc20SupportT>: TryFrom<U256> + EvmData,
```
MOZGIII, line 122:
```suggestion
    AllowanceOf<Erc20SupportT>: TryFrom<U256>,
```

Doesn't seem like it's needed here, can we drop it?

dmitrylavrenov: It's needed here - https://github.com/humanode-network/humanode/blob/e9d0e8cfadfb50cd099611d55745e195fcd2d266/crates/precompile-native-currency/src/lib.rs#L187.

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: Redesigned based on discussion - https://github.com/humanode-network/humanode/pull/809/commits/e87a191a4722ca14acf015006cb8044b456b50a6

crates/pallet-erc20-support/src/migrations/v1.rs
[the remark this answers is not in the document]
dmitrylavrenov: Redesigned based on discussion - https://github.com/humanode-network/humanode/pull/809/commits/e87a191a4722ca14acf015006cb8044b456b50a6

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: Redesigned based on discussion - https://github.com/humanode-network/humanode/pull/809/commits/e87a191a4722ca14acf015006cb8044b456b50a6

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/d2a676618c492874d6de82720816d12193a4b02a

crates/precompile-native-currency/src/lib.rs
MOZGIII: Ah, got it, thanks for the clarification!

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/db62ae49b66b28cc6e7afc69d6b9ac0e068c43f5

MOZGIII: We can remove this fn altogether

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/cd5369271d3d5983bf113f7a46ff47679d54b4d1

MOZGIII: The issue arises with how one would implement the `post_migrate` calls, given that we how to determine whether to run the migration or not at the call layer. I think in this sense we should only run the `post_migrate` for the *latest* migration, or implement more "atomic" verification calls that could be used in a non-trivial way at the `post_upgrade` level.

crates/pallet-erc20-support/src/migrations/v1.rs
```diff
@@ -0,0 +1,20 @@
...
+use frame_support::{sp_tracing::info, traits::Get, weights::Weight};
+
+use crate::BalanceOf;
+use crate::{Approvals, Config};
+
+/// Migrate from version 0 to 1.
+pub fn migrate<T: Config<I>, I: 'static>() -> Weight {
+    info!("Running migration to v1");
+
+    let mut weight: Weight = T::DbWeight::get().reads(0);
```
MOZGIII, line 12: `reads(0)`?

crates/pallet-erc20-support/src/lib.rs
```diff
@@ -89,6 +105,33 @@ pub mod pallet {
...
+        fn on_runtime_upgrade() -> Weight {
+            let mut weight = T::DbWeight::get().reads(1);
+
+            if StorageVersion::get::<Pallet<T, I>>() == 0 {
+                weight.saturating_accrue(migrations::v1::migrate::<T, I>());
+                StorageVersion::new(1).put::<Pallet<T, I>>();
+                weight.saturating_accrue(T::DbWeight::get().writes(1));
+            }
+
+            assert_eq!(<Pallet<T, I>>::on_chain_storage_version(), STORAGE_VERSION);
```
MOZGIII, line 120: I meant for this to be at the `post_upgrade` check

MOZGIII: Also don't use `STORAGE_VERSION` directly, use `<Pallet<T, I>>::current_storage_version()` instead

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: Yep, as we should start from zero.

crates/pallet-erc20-support/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/6a5072127188411bfea8fce855652917d1553d25

crates/pallet-erc20-support/src/migrations/v1.rs
MOZGIII: This could be better expressed with something like `::new()` or `::default()` or `::zero()` (not sure which ones are available).

crates/pallet-erc20-support/src/migrations/v1.rs
dmitrylavrenov: Thanks, I got your thoughts.

crates/pallet-erc20-support/src/migrations/v1.rs
MOZGIII: Actually, just `Wight::new`

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/809/commits/23c0d22b16e810e4bb13d71a3005000968b8df8d

MOZGIII (approved):

PR #867, opened 2024-01-12
Support for running the node with manual/instant seal consensus

### LINKED ISSUE #773: Support for running the node with instant seal consensus
The experience of developing smart contracts against our chain can be significantly improved if we add an instant seal consensus operation mode. The idea is the node will be sealing the blocks automatically as soon as there is anything to include in the block instead of going through the usual BABE/Grandpa flow. This obviously means that nodes in this mode will be unable to do any sort of networked consensus - but this is fine since this is intended only for single-node development node, where a developer wants to test their code in an environment that sort of resembles the real chain operation but is much quicker (i.e. for running the tests in a TDD flow).

The node running in this mode will have some obvious shortcomings, like
- timestamps that are way closer to each other than in the real chain (may be was less than a second, while the in the real world blocks are typically 6 seconds apart)
- different block seals (not a concern for the EVM developers most likely, but it is a notable difference)
- possibly, other small niche differences

dmitrylavrenov:
## instant sealing
```
2024-01-12 18:26:26 💤 Idle (0 peers), best: #0 (0xbdd3…5a12), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:26:31 💤 Idle (0 peers), best: #0 (0xbdd3…5a12), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:26:36 💤 Idle (0 peers), best: #0 (0xbdd3…5a12), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:26:37 🙌 Starting consensus session on top of parent 0xbdd3ceaea17ba0d7e44f17f68989c659393c96ed2b79ceab18612be0dfd95a12    
2024-01-12 18:26:37 🎁 Prepared block for proposing at 1 (15 ms) [hash: 0xb27f29910cb1a425ae7d854091b24c3408fbdea1e08ac49cd102c7ccbc82a6f4; parent_hash: 0xbdd3…5a12; extrinsics (2): [0x3b92…0797, 0x20c7…a973]]    
2024-01-12 18:26:37 ✨ Imported #1 (0xb27f…a6f4)    
2024-01-12 18:26:37 Instant Seal success: CreatedBlock { hash: 0xb27f29910cb1a425ae7d854091b24c3408fbdea1e08ac49cd102c7ccbc82a6f4, aux: ImportedAux { header_only: false, clear_justification_requests: false, needs_justification: false, bad_justification: false, is_new_best: true } }    
2024-01-12 18:26:41 💤 Idle (0 peers), best: #1 (0xb27f…a6f4), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:26:46 💤 Idle (0 peers), best: #1 (0xb27f…a6f4), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0 
```
## manual sealing
```
➜  humanode git:(consensus-manual-sealing) curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "engine_createBlock", "params": [true, false, null]}' http://127.0.0.1:9933
{"jsonrpc":"2.0","result":{"hash":"0x1d63fd925aab318dfdeb293dbb42736105fc00b4a29de00690d3f5188b40566d","aux":{"header_only":false,"clear_justification_requests":false,"needs_justification":false,"bad_justification":false,"is_new_best":true}},"id":1}% 
```

```
2024-01-12 18:45:19 Accepting new connection 1/16000
2024-01-12 18:45:19 🙌 Starting consensus session on top of parent 0xbdd3ceaea17ba0d7e44f17f68989c659393c96ed2b79ceab18612be0dfd95a12    
2024-01-12 18:45:19 🎁 Prepared block for proposing at 1 (5 ms) [hash: 0x1d63fd925aab318dfdeb293dbb42736105fc00b4a29de00690d3f5188b40566d; parent_hash: 0xbdd3…5a12; extrinsics (1): [0x3b92…0797]]    
2024-01-12 18:45:19 ✨ Imported #1 (0x1d63…566d)    
2024-01-12 18:45:23 💤 Idle (0 peers), best: #1 (0x1d63…566d), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:45:28 💤 Idle (0 peers), best: #1 (0x1d63…566d), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0    
2024-01-12 18:45:33 💤 Idle (0 peers), best: #1 (0x1d63…566d), finalized #0 (0xbdd3…5a12), ⬇ 0 ⬆ 0
```

### DIFF
--- a/crates/humanode-peer/src/chain_spec.rs
+++ b/crates/humanode-peer/src/chain_spec.rs
@@ -6,7 +6,7 @@ use hex_literal::hex;
 use humanode_runtime::{
     opaque::SessionKeys, robonode, token_claims::types::ClaimInfo, AccountId, BabeConfig, Balance,
     BalancesConfig, BioauthConfig, BootnodesConfig, ChainPropertiesConfig, EVMConfig,
-    EnableManualSeal, EthereumAddress, EthereumChainIdConfig, EthereumConfig, EvmAccountId,
+    EnableDevBlockImportSeal, EthereumAddress, EthereumChainIdConfig, EthereumConfig, EvmAccountId,
     GenesisConfig, GrandpaConfig, ImOnlineConfig, SessionConfig, Signature, SudoConfig,
     SystemConfig, TokenClaimsConfig, WASM_BINARY,
 };
@@ -29,15 +29,15 @@ pub type DevChainSpec = sc_service::GenericChainSpec<DevGenesisConfig, Extension
 pub struct DevGenesisConfig {
     /// Genesis config.
     genesis_config: GenesisConfig,
-    /// The flag that if enable manual-seal mode.
-    enable_manual_seal: Option<bool>,
+    /// The flag that if enable block import manual seal mode.
+    enable_block_import_manual_seal: Option<bool>,
 }
 
 impl sp_runtime::BuildStorage for DevGenesisConfig {
     fn assimilate_storage(&self, storage: &mut Storage) -> Result<(), String> {
         BasicExternalities::execute_with_storage(storage, || {
-            if let Some(enable_manual_seal) = &self.enable_manual_seal {
-                EnableManualSeal::set(enable_manual_seal);
+            if let Some(enable_block_import_manual_seal) = &self.enable_block_import_manual_seal {
+                EnableDevBlockImportSeal::set(enable_block_import_manual_seal);
             }
         });
         self.genesis_config.assimilate_storage(storage)
@@ -173,7 +173,9 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
 }
 
 /// A configuration for dev.
-pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSpec, String> {
+pub fn development_config(
+    enable_block_import_manual_seal: Option<bool>,
+) -> Result<DevChainSpec, String> {
     let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?;
 
     let robonode_public_key = dev_robonode_public_key(&DEFAULT_DEV_ROBONODE_PUBLIC_KEY)?;
@@ -206,7 +208,7 @@ pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSp
                     robonode_public_key,
                     vec![account_id("Alice")],
                 ),
-                enable_manual_seal,
+                enable_block_import_manual_seal,
             }
         },
         // Bootnodes
--- a/crates/humanode-peer/src/chain_spec.rs
+++ b/crates/humanode-peer/src/chain_spec.rs
@@ -431,15 +431,9 @@ mod tests {
     use super::*;
 
     fn assert_genesis_config(
-        chain_spec_result: Result<ChainSpec, String>,
+        chain_spec_result: Result<impl BuildStorage, String>,
     ) -> sp_core::storage::Storage {
         chain_spec_result.unwrap().build_storage().unwrap()
-    }
-
-    fn assert_dev_genesis_config(
-        dev_chain_spec_result: Result<DevChainSpec, String>,
-    ) -> sp_core::storage::Storage {
-        dev_chain_spec_result.unwrap().build_storage().unwrap()
     }
 
     fn assert_balanced_currency_swap(storage: sp_core::storage::Storage) {
@@ -458,7 +452,7 @@ mod tests {
 
     #[test]
     fn development_config_works() {
-        let storage = assert_dev_genesis_config(development_config(None));
+        let storage = assert_genesis_config(development_config(None));
         assert_balanced_currency_swap(storage);
     }
 
--- a/crates/humanode-peer/src/chain_spec.rs
+++ b/crates/humanode-peer/src/chain_spec.rs
@@ -1,14 +1,14 @@
 //! Provides the [`ChainSpec`] portion of the config.
 
 use crypto_utils::{authority_keys_from_seed, get_account_id_from_seed};
-use frame_support::BoundedVec;
+use frame_support::{BasicExternalities, BoundedVec};
 use hex_literal::hex;
 use humanode_runtime::{
     opaque::SessionKeys, robonode, token_claims::types::ClaimInfo, AccountId, BabeConfig, Balance,
     BalancesConfig, BioauthConfig, BootnodesConfig, ChainPropertiesConfig, EVMConfig,
-    EthereumAddress, EthereumChainIdConfig, EthereumConfig, EvmAccountId, GenesisConfig,
-    GrandpaConfig, ImOnlineConfig, SessionConfig, Signature, SudoConfig, SystemConfig,
-    TokenClaimsConfig, WASM_BINARY,
+    EnableManualSeal, EthereumAddress, EthereumChainIdConfig, EthereumConfig, EvmAccountId,
+    GenesisConfig, GrandpaConfig, ImOnlineConfig, SessionConfig, Signature, SudoConfig,
+    SystemConfig, TokenClaimsConfig, WASM_BINARY,
 };
 use pallet_im_online::sr25519::AuthorityId as ImOnlineId;
 use sc_chain_spec_derive::{ChainSpecExtension, ChainSpecGroup};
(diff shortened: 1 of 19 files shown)


crates/humanode-peer/src/cli/root.rs
```diff
@@ -44,7 +44,13 @@ impl SubstrateCli for Root {
 
     fn load_spec(&self, id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
         Ok(match id {
-            "dev" => Box::new(chain_spec::development_config()?),
+            "dev" => {
+                let enable_dev_block_import_seal =
+                    self.run.dev_block_import_sealing.clone().map(|_| true);
```
MOZGIII, line 49:
```suggestion
                let enable_dev_block_import_seal =
                    self.run.dev_block_import_sealing.is_some();
```

crates/humanode-peer/src/service/mod.rs
```diff
@@ -329,6 +336,9 @@ pub async fn new_full(config: Configuration) -> Result<TaskManager, ServiceError
         reqwest: reqwest::Client::new(),
     });
 
+    // Channel for the rpc handler to communicate with the authorship task.
+    let (command_sink, commands_stream) = futures::channel::mpsc::channel(1000);
```
MOZGIII, line 340: We probably shouldn't create this if we don't use it

crates/humanode-runtime/src/lib.rs
```diff
@@ -385,9 +385,23 @@ impl pallet_grandpa::Config for Runtime {
...
+
+pub struct ConsensusOnTimestampSet<T>(PhantomData<T>);
+impl<T: pallet_babe::Config> OnTimestampSet<T::Moment> for ConsensusOnTimestampSet<T> {
+    fn on_timestamp_set(moment: T::Moment) {
+        if EnableDevBlockImportSeal::get() {
+            return;
+        }
+        <pallet_babe::Pallet<T> as OnTimestampSet<T::Moment>>::on_timestamp_set(moment)
+    }
+}
```
MOZGIII, line 400: Interesting, why is this needed?

crates/humanode-peer/src/cli/config.rs
```diff
@@ -67,11 +67,14 @@ pub trait CliConfigurationExt: SubstrateCliConfigurationProvider {
                 })
         });
 
+        let dev_block_import_sealing = self.dev_block_import_sealing().unwrap().clone();
```
MOZGIII, line 70: `.unwrap()` here

crates/humanode-rpc/src/lib.rs
```diff
@@ -12,6 +12,7 @@ use fc_rpc::{
 };
 use fc_rpc_core::types::{FeeHistoryCache, FilterPool};
 use fc_storage::OverrideHandle;
+use futures::channel::mpsc;
```
MOZGIII, line 15: Why not `tokio`?

MOZGIII: I get it - it is an explicit type requirement of the implementation

crates/humanode-rpc/src/lib.rs
```diff
@@ -146,6 +148,8 @@ pub struct Deps<C, P, BE, VKE, VSF, A: ChainApi, SC> {
     pub evm: EvmDeps,
     /// Subscription task executor instance.
     pub subscription_task_executor: sc_rpc::SubscriptionTaskExecutor,
+    /// Manual seal command sink.
+    pub command_sink: Option<mpsc::Sender<EngineCommand<Hash>>>,
```
MOZGIII, line 152:
```suggestion
    pub manual_seal_command_tx: Option<mpsc::Sender<EngineCommand<Hash>>>,
```

crates/humanode-peer/src/service/mod.rs
```diff
@@ -21,11 +20,13 @@ use sc_service::{
     Error as ServiceError, KeystoreContainer, PartialComponents, TaskManager, WarpSyncParams,
 };
 use sc_telemetry::{Telemetry, TelemetryWorker};
+use sp_api::TransactionFor;
 use tracing::*;
 
-use crate::configuration::Configuration;
+use crate::{cli::DevBlockImportSealing, configuration::Configuration};
```
MOZGIII, line 26: Ideally service should not have dependencies on the cli

crates/humanode-peer/src/service/import_queue.rs
```diff
@@ -0,0 +1,81 @@
...
+    client: Arc<FullClient>,
+    config: &sc_service::Configuration,
+    task_manager: &TaskManager,
+    frontier_backend: Arc<FrontierBackend>,
+) -> (
+    sc_consensus::DefaultImportQueue<Block, FullClient>,
+    FullBoxBlockImport,
+) {
+    let block_import = FrontierBlockImport::new(
+        Arc::clone(&client),
```
MOZGIII, line 26: The first argument here is usually `babe_block_import` - why don't we also pass it here? It is available at the callsite at the `service:: new_partial`, so passing is easy - so I assume there must be a reason. Also, if there is a reason - why do we still needlessly create the babe block import at the callsite of this `sealing` fn at service?

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -150,38 +173,43 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
 }
 
 /// A configuration for dev.
-pub fn development_config() -> Result<ChainSpec, String> {
+pub fn development_config(
+    enable_block_import_manual_seal: Option<bool>,
```
MOZGIII, line 177:
```suggestion
    enable_block_import_manual_seal: bool,
```

crates/humanode-peer/src/chain_spec.rs
```diff
@@ -1,26 +1,49 @@
...
+/// Specialized `ChainSpec` for development.
+pub type DevChainSpec = sc_service::GenericChainSpec<DevGenesisConfig, Extensions>;
+
+/// Extension for the dev genesis config to support a custom changes to the genesis state.
+#[derive(Serialize, Deserialize)]
+pub struct DevGenesisConfig {
+    /// Genesis config.
+    genesis_config: GenesisConfig,
+    /// The flag that if enable block import manual seal mode.
+    enable_block_import_manual_seal: Option<bool>,
```
MOZGIII, line 33:
```suggestion
    enable_block_import_manual_seal: bool,
```
We can just use a boolean here and set to true if it has to be true.

This can also be used to embed arbitrary robonode key into the runtime, i.e. for benchmarks.

But then again - why not create a full-blown pallet for this? If we need a storage and and genesis config - why bother and do this magic?

crates/humanode-runtime/src/lib.rs
```diff
@@ -385,9 +385,23 @@ impl pallet_grandpa::Config for Runtime {
 /// A timestamp: milliseconds since the unix epoch.
 pub type UnixMilliseconds = u64;
 
+parameter_types! {
+    pub storage EnableDevBlockImportSeal: bool = false;
```
MOZGIII, line 389: Oh, cool, I didn't know we can use the `storage` at `parameter_types!`

crates/humanode-runtime/src/lib.rs
MOZGIII: Do we maybe want to have a second, different runtime for the instant/manual seal consensus, without babe and grandpa pallets/APIs?

crates/humanode-peer/src/service/inherents.rs
```diff
@@ -104,3 +104,39 @@ impl<Client> Clone for Creator<Client> {
         }
     }
 }
+
+/// Sealing case related inherent data providers.
+pub mod sealing {
+    use std::cell::RefCell;
+
+    thread_local!(static TIMESTAMP: RefCell<u64> = RefCell::new(0));
```
MOZGIII, line 112: You can't use thread local values in the async context, because async fns can be scheduled to the async runtime as tasks, which means they'll execute on a thread pool, meaning each thread will have its own value - and the values will not be shared. If you need a local variable - just use a static variable with an access serialization method like mutex, or an atomic (since it is a primitive integer).

crates/humanode-peer/src/service/inherents.rs
```diff
@@ -104,3 +104,39 @@ impl<Client> Clone for Creator<Client> {
...
+            &self,
+            inherent_data: &mut sp_inherents::InherentData,
+        ) -> Result<(), sp_inherents::Error> {
+            TIMESTAMP.with(|x| {
+                let current_timestamp = *x.borrow();
+                *x.borrow_mut() = current_timestamp
+                    .checked_add(humanode_runtime::SLOT_DURATION)
+                    .expect("operation is under control, sealing case is allowed only in dev mode");
+                inherent_data.put_data(sp_timestamp::INHERENT_IDENTIFIER, &*x.borrow())
+            })
```
MOZGIII, line 130: This code assumes `provide_inherent_data` is called once for each block. I am not sure the code we have makes any effort to guarantee this - so no reason why this should be true. It seems like we need to do something else here.

dmitrylavrenov:
Discussed and agreed to have separate runtime for running the node with manual/instant seal consensus.

PR #914, opened 2024-02-15
Add a new unified Dockerfile and docker-bake.hcl

MOZGIII:


MOZGIII:
This performs so much worse in practice D:

PR #1005, opened 2024-04-17
Use `--checks=all` at e2e `on-runtime-upgrade` try-runtime test

dmitrylavrenov:


crates/pallet-humanode-session/src/migrations/v1.rs
```diff
@@ -64,14 +64,21 @@ pub fn migrate<T: Config>() -> Weight {
...
+        // Record the count of identities.
+        let identities_count: u64 = <CurrentSessionIdentities<T>>::iter()
+            .count()
+            .try_into()
+            .unwrap();
+        return identities_count.encode();
+    }
+
+    let identities_count: u64 = <SessionIdentities<T>>::iter().count().try_into().unwrap();
     identities_count.encode()
```
MOZGIII, line 82: Let's be more explicit here: disable the check and data population entirely for newer versions. Otherwise we might get annoyed by this again in the future.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1005/commits/6b4afa33e59239e9214cc538e0da3801aee67b82

MOZGIII: At the `post_migrate` check, run the assertions depending on whether the `state` is empty or not. At the `pre_migrate` - only write the `state` if needed. Also, we should make sure the state is non-empty when we encode the `0` identities could - to distinguish between the empty state indicating no checks and the migration of 0 entries.

crates/pallet-humanode-session/src/migrations/v1.rs
```diff
@@ -64,14 +64,7 @@ pub fn migrate<T: Config>() -> Weight {
 /// Panics if anything goes wrong.
 #[cfg(feature = "try-runtime")]
 pub fn pre_migrate<T: Config>() -> Vec<u8> {
-    // Ensure the new identities don't exist yet (i.e. we have clear space to migrate).
-    assert_eq!(<SessionIdentities<T>>::iter().next(), None);
```
MOZGIII, line 68: No, not drop it completely :D

dmitrylavrenov: Got it, reimplemented.

crates/pallet-humanode-session/src/migrations/v1.rs
```diff
@@ -64,6 +64,13 @@ pub fn migrate<T: Config>() -> Weight {
 /// Panics if anything goes wrong.
 #[cfg(feature = "try-runtime")]
 pub fn pre_migrate<T: Config>() -> Vec<u8> {
+    let onchain = <Pallet<T>>::on_chain_storage_version();
+
+    // Disable the check for newer versions by returning an empty state.
+    if onchain > 0 {
```
MOZGIII, line 70:
```suggestion
    if onchain >= 1 {
```

crates/pallet-humanode-session/src/migrations/v1.rs
```diff
@@ -80,6 +87,11 @@ pub fn pre_migrate<T: Config>() -> Vec<u8> {
 /// Panics if anything goes wrong.
 #[cfg(feature = "try-runtime")]
 pub fn post_migrate<T: Config>(state: Vec<u8>) {
+    // Empty state means that the check is disabled.
+    if state.is_empty() {
```
MOZGIII, line 91: Isn't `0u64.encode().is_empty() == true` as well?

dmitrylavrenov: Will try

dmitrylavrenov: Tried `return 0u64.encode();` and got failed assertion.

MOZGIII: So it's non-empty? Well then

PR #1034, opened 2024-05-09
Add snapshot tests for other subcommands `--help` output

### LINKED ISSUE #1000: Add snapshot tests for other subcommands `--help` output

dmitrylavrenov:
Closes #1000


MOZGIII (commented):
We probably want to place those fixtures into a subdir

dmitrylavrenov:
> We probably want to place those fixtures into a subdir

Reimplemented a little bit and moved to separate subdirectories all help related commands.

utils/e2e-tests/bash/tests/base/help.sh
```diff
@@ -3,23 +3,44 @@ set -euo pipefail
...
+    -I '^humanode-peer {{sha}}$'        # template
+    -I '^humanode-peer [0-9a-f]\{40\}$' # expected
+  )
+
+  if ! DIFF="$(diff "${DIFF_CMD_ARGS[@]}" <(printf '%s' "$TEMPLATE") <(printf '%s' "$OUTPUT"))"; then
+    printf "Output did not match:\n%s\n" "$DIFF"
+    exit 1
+  fi
+
+  local SUBCOMMANDS=($(awk '/Commands:/,/Options:/{if(/Commands:|Options:/) next; print}' <(printf '%s' "$OUTPUT") | awk '{if ($1) print $1}'))
```
MOZGIII, line 33: Why the `awk` command chaining here?

It seems to me like the `awk` invocation here would look better if we extract it into a separate fn.

Same applied to the `if`/`diff` clause.

I'd refactor this a bit further to simplify the subsequent maintenance.

MOZGIII: Also, there are a ton of shellcheck warnings:
image

MOZGIII: Refactored a lot of stuff, but still left the chained `awk` invocations as is for you to look into.

dmitrylavrenov: Thanks for the improvements!

The `awk` usage has been refactored - https://github.com/humanode-network/humanode/pull/1034/commits/66cfa4912f927af83ab4796e80906785159978b6.

MOZGIII (approved):

PR #1026, opened 2024-05-07
Integrate `scanResultBlob` pattern on `robonode-client` and `bioauth-flow-rpc` 

dmitrylavrenov:
Should be merged after #1065 

As agreed integration of `scanResultBlob` pattern on `bioauth-flow-rpc` and `robonode-client` sides should be backward compatible.

PR #1037, opened 2024-05-13
Integrate `scanResultBlob` pattern on robonode side

dmitrylavrenov:
Agreed on decoupling logic of #1026


dmitrylavrenov:
Agreed on decoupling logic of #1026

MOZGIII (approved):

MOZGIII (changes requested):
Actually hold on

MOZGIII (changes requested):
Add tests to ensure errors return `scan_result_blob` too

crates/robonode-server/src/logic/mod.rs
```diff
@@ -20,6 +20,9 @@ pub mod traits;
 
 pub use traits::*;
 
+/// A type alias representing a scan result blob.
+pub(crate) type ScanResultBlob = Option<String>;
```
MOZGIII, line 24:
```suggestion
pub(crate) type ScanResultBlob = String;
```

Applying option like this is very confusing when it comes to reading the code, better apply it at the usage location: `Option<ScanResultBlob>`

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1037/commits/33f49d52f6000f24f1c3ffb850fa4a23a8e86cd7

dmitrylavrenov:
Closes in favour of #1065

PR #1065, opened 2024-06-20
Integrate `scanResultBlob` pattern on `facetec-api-client` and `robonode-server`

dmitrylavrenov:
As agreed integration of `scanResultBlob` pattern on `facetec-api-client` and `robonode-server` sides: 
- backward compatibility without needs to be supported
- `scanResultBlob` data is always expected in responses from `facetec-server`.


crates/robonode-server/src/logic/op_authenticate.rs
```diff
@@ -140,39 +146,47 @@ where
...
+                scan_result_blob,
+            ));
         }
 
-        let public_key_bytes =
-            hex::decode(&found.identifier).map_err(|_| Error::InternalErrorInvalidPublicKeyHex)?;
-        let public_key =
-            PK::try_from(&public_key_bytes).map_err(|_| Error::InternalErrorInvalidPublicKey)?;
+        let public_key_bytes = hex::decode(&found.identifier)
+            .map_err(|_| Error::InternalErrorInvalidPublicKeyHex(scan_result_blob.clone()))?;
```
MOZGIII, line 177: Use `match` to avoid a clone here and in similar locations

MOZGIII (approved):
Looks good, however I suggest that we finalize this and run the flow end-to-end at least once before merging

crates/robonode-server/src/logic/op_authenticate.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1065/commits/461f645870a1df236da551e5a596424153e99a6e

MOZGIII (approved):
The code looks good!

Is it backwards-compatible with the currently deployed nodes? If so we can merge this I think.

crates/robonode-server/src/logic/op_authenticate.rs
```diff
@@ -143,39 +149,59 @@ where
...
             }
         };
 
         // If the results set is empty - this means that this person was not
         // found in the system.
-        let found = results.first().ok_or(Error::PersonNotFound)?;
+        let found = match results.first() {
+            Some(found) => found,
+            None => return Err(Error::PersonNotFound(scan_result_blob)),
+        };
```
MOZGIII, line 173:
```suggestion
        let Some(found) = results.first() else {
            return Err(Error::PersonNotFound(scan_result_blob));
        };
```

dmitrylavrenov:
> Looks good, however I suggest that we finalize this and run the flow end-to-end at least once before merging

As I remember, we've agreed on that `facetec-api-client` and `robonode-server` don't need to support backward-compatibility. Should we change this decision now ?

MOZGIII:
> Should we change this decision now ?

No, not necessarily.



> The description of this PR contains agreements we've set last time we discussed it.

Let's just discuss them in more detail, since as they are written they leave some possibility of issues; we can just rewrite them to be more precise after this.

MOZGIII:
To sum up our discussion in chat:

- we do not need compatibility at the `facetec-api-client` - we always deal with a known version of the server there
- we do need compatibility between the old `robonode-client` and new `robonode-server` - as nodes may be arbitrarily old, and we'd want them to still work with the new robonode server as long as the rest of the subsystems also allow it

MOZGIII (approved):

PR #1248, opened 2024-08-29
Support for disabling validators via sudo [option 1]

dmitrylavrenov:


dmitrylavrenov:
Closes in favour of #1249

PR #1249, opened 2024-08-29
Support for banning validators via sudo at `pallet-humanode-session`

dmitrylavrenov:


crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +71,20 @@ pub mod pallet {
...
 
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        /// Disable validator.
+        #[pallet::call_index(0)]
+        #[pallet::weight(0)]
+        pub fn disable(origin: OriginFor<T>, validator_id: T::ValidatorId) -> DispatchResult {
+            ensure_root(origin)?;
+
+            <pallet_session::Pallet<T>>::disable(&validator_id);
```
MOZGIII, line 82: Disallow bootnodes here

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +74,67 @@ pub mod pallet {
...
+
+            Self::disable(account_id)?;
+
+            Ok(())
+        }
+
+        /// Ban account.
+        #[pallet::call_index(1)]
+        #[pallet::weight(0)]
+        pub fn ban(origin: OriginFor<T>, account_id: T::AccountId) -> DispatchResult {
```
MOZGIII, line 114: Add `unban`?

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +74,67 @@ pub mod pallet {
...
+        TooManyBannedAccounts,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        /// Kick validator based on provided account id.
+        #[pallet::call_index(0)]
+        #[pallet::weight(0)]
+        pub fn kick(origin: OriginFor<T>, account_id: T::AccountId) -> DispatchResult {
+            ensure_root(origin)?;
```
MOZGIII, line 104: Can we make this generic, so that we can tweak this check at the runtime level?

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: Added, ah, sorry, forgot to push :)

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +74,67 @@ pub mod pallet {
...
+
+            ensure!(
+                !Self::is_banned(&account_id),
+                Error::<T>::AccountIsAlreadyBanned
+            );
+
+            with_storage_layer(move || {
+                Self::disable(account_id.clone())?;
+
+                <BannedAccounts<T>>::try_mutate::<_, DispatchError, _>(move |banned_accounts| {
```
MOZGIII, line 125: Why not try_push directly?

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -29,7 +29,6 @@ pub mod pallet {
     #[pallet::config]
     pub trait Config:
         frame_system::Config
-        /* Session pallet is only required for migration to v1. */
```
MOZGIII, line 32: Why remove this?

MOZGIII (commented):
Overall, this strategy seems great!

I think we should move the banlist management to an external crate, as it seems we really only need an `is_banned` for interfacing between the two; also, we should make it so ban does not directly exclude the validator from the list. I have doubts still about this, so maybe this is fine.

Also, there is one interesting interaction there that I think was overlooked: when someone is kicked and banned, they will from out of the current validators set, but if they were in the QueuedKeys they'll still become a validator on the next session, no?

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: As we need to use `T::ValidatorIdOf` as a part of `frame_session::Config` to be passed to `<pallet_session::Pallet<T>>::disable(..)`.

So, it's not just used for migration.

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: Thanks, good suggestions - https://github.com/humanode-network/humanode/pull/1249/commits/c85780364ab73b6932cb581d829192e0fb5e7ebc.

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: So, it can be executed not just as sudo call?

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: Not removed now.

crates/pallet-humanode-session/src/lib.rs
dmitrylavrenov: Kick call is removed.

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +75,66 @@ pub mod pallet {
...
 
+    /// A list of all banned accounts that can't be validators in the network.
+    #[pallet::storage]
+    pub type BannedAccounts<T: Config> =
+        StorageValue<_, BoundedVec<T::AccountId, T::MaxBannedAccounts>, ValueQuery>;
+
+    /// Possible errors.
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Attempt to ban bootnode to be a validator.
```
MOZGIII, line 86:
```suggestion
        /// Banning bootnodes is not allowed.
```

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +75,66 @@ pub mod pallet {
     #[pallet::storage]
     pub type CurrentSessionIndex<T: Config> = StorageValue<_, SessionIndex, OptionQuery>;
 
+    /// A list of all banned accounts that can't be validators in the network.
+    #[pallet::storage]
+    pub type BannedAccounts<T: Config> =
+        StorageValue<_, BoundedVec<T::AccountId, T::MaxBannedAccounts>, ValueQuery>;
```
MOZGIII, line 81: Our use of the storage item indicates it would be much better as a set instead of a vec

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1249/commits/5f9e7bbb234e3a3192de8a7a45475cb77fa26043

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -72,6 +76,73 @@ pub mod pallet {
     #[pallet::storage]
     pub type CurrentSessionIndex<T: Config> = StorageValue<_, SessionIndex, OptionQuery>;
 
+    /// A set of all banned accounts that can't be validators in the network.
+    #[pallet::storage]
+    pub type BannedAccounts<T: Config> =
+        StorageValue<_, BoundedBTreeSet<T::AccountId, T::MaxBannedAccounts>, ValueQuery>;
```
MOZGIII, line 82: But why?..

Use `StorageMap`?

MOZGIII: Either do that, or drop the `is_banned` API and replace with something that operates on the entire value - as `is_banned` will pay the price of load the whole thing while only using part of it; we'd be better of with loading the whole thing and running the filtering over with maintaining the same access, instead of dropping and re-acquiring the ref (there is storage cache, but we'd still pay for the deconding/enoding as I understand).

Or, just use map if only the access to individual keys is anticipated (which might be less optimal here).

dmitrylavrenov: Good point, but I would prefer to user `StorageValue` with `BTreeSet` as it should be more optimal here. But, sure, we should avoid dropping and re-acquiring the ref, reimplemented https://github.com/humanode-network/humanode/pull/1249/commits/cfeb9204f445c07105bc7f397db3b0c4b5bb45c5, `is_banned` API has been dropped.

crates/pallet-humanode-session/src/lib.rs
```diff
@@ -138,13 +227,16 @@ impl<T: Config> Pallet<T> {
...
+
         let bioauth_active_authentications = <pallet_bioauth::Pallet<T>>::active_authentications()
             .into_inner()
             .into_iter()
             .take(T::MaxBioauthValidators::get().try_into().unwrap())
-            .filter_map(|authentication| {
+            .filter_map(move |authentication| {
                 T::ValidatorPublicKeyOf::convert(authentication.public_key.clone())
                     .map(|account_id| (account_id, Identification::Bioauth(authentication)))
+                    .filter(|(account_id, _)| !banned_accounts.contains(account_id))
```
MOZGIII, line 239:
```suggestion
                    .filter(|account_id| !banned_accounts.contains(account_id))
                    .map(|account_id| (account_id, Identification::Bioauth(authentication)))
```

crates/humanode-runtime/src/weights/pallet_humanode_session.rs
MOZGIII, line 1: Computed weights may be inconsistent with the rest of the computed weights.

MOZGIII (approved):

crates/humanode-runtime/src/weights/pallet_humanode_session.rs
dmitrylavrenov: Got it, will regenerate all weights.

dmitrylavrenov: Done

MOZGIII (approved):

PR #1253, opened 2024-09-03
Support for kicking validators via sudo at `pallet-humanode-session`

dmitrylavrenov:


dmitrylavrenov:
Looks like it's not safe to use `disable` - https://github.com/paritytech/polkadot-sdk/issues/475#issuecomment-1691724642

PR #1268, opened 2024-09-13
Add more controlled accounts for balances, vesting, token claims logic at `patch-forked-spec` util

dmitrylavrenov:


MOZGIII (changes requested):
This is does not follow the previous logic for construction of the values;

Overall, I'm not too pleased with this solution conceptually - we might want to find another way.

PR #1291, opened 2024-10-09
Move the bioauth OffenceSlasher from the runtime to its own pallet

### LINKED ISSUE #787: Move the bioauth `OffenceSlasher` from the runtime to its own pallet
The `OffenceSlasher` type in the runtime implements the `sp_staking::offence::OnOffenceHandler` in a way that terminates the user's bioauth.

This is not too complicated, but we still should move it into a dedicated pallet and implement a suite of proper tests for it.

For some context, it is not a part of the `pallet_bioauth` because it would pull in `sp_staking` as a dependency - something we don't want/need there. So, to avoid cluttering the `pallet_bioauth` implementation we had to push it elsewhere - to the runtime in this case This is not good enough though, and it is time to introduce its own place for the offence handling logic.

dmitrylavrenov:
Closes #787


crates/humanode-runtime/src/lib.rs
```diff
@@ -568,64 +570,12 @@ impl pallet_humanode_session::Config for Runtime {
...
-    }
-}
-
 impl pallet_im_online::Config for Runtime {
     type AuthorityId = ImOnlineId;
     type RuntimeEvent = RuntimeEvent;
     type NextSessionRotation = Babe;
     type ValidatorSet = Historical;
-    type ReportUnresponsiveness = Offences;
+    type ReportUnresponsiveness = OffenceHandler<Offences>;
```
MOZGIII, line 578: Do we need pallet offences at all?

dmitrylavrenov: Good question, we don't use any logic provided by this pallet. So, we can easily remove it now with own offence handler, I think.

dmitrylavrenov: Done

crates/pallet-humanode-offences/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        T::OnOffence::on_offence(offenders);
+
+        Ok(())
+    }
+
+    fn is_known_offence(_offenders: &[T::Offender], _time_slot: &O::TimeSlot) -> bool {
+        false
+    }
+}
```
MOZGIII, line 55: That's not very useful on its own, although very flexible as it delegates the entirety of its implementation to the `OnOffence` handler...

How about we implement the slashing logic right here instead?

Cargo.lock
```diff
@@ -260,9 +260,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.91"
+version = "1.0.89"
```
MOZGIII, line 263: An odd downgrade

MOZGIII: :D

crates/pallet-humanode-offences/src/lib.rs
dmitrylavrenov: Implemented

MOZGIII:
> 2024-11-12 16:50:48 Removed 59778 Offences keys 🧹

I haven't read the code just yet, but we don't want to do a synchronous deletion of the keys as it can exhaust the block size and fail. We need to first set up the facilities for making async collection of garbage, and then use them; or, alternatively, don't do it at the rumtime at all, and remove the keys manually instead via sudo txs - that I think can be done even today without any extra development.

dmitrylavrenov:
> > 2024-11-12 16:50:48 Removed 59778 Offences keys 🧹
> 
> I haven't read the code just yet, but we don't want to do a synchronous deletion of the keys as it can exhaust the block size and fail. We need to first set up the facilities for making async collection of garbage, and then use them; or, alternatively, don't do it at the rumtime at all, and remove the keys manually instead via sudo txs - that I think can be done even today without any extra development.

Got it, will remove the code related to deleting keys at the runtime level during runtime upgrade execution.

crates/pallet-humanode-offences/src/lib.rs
MOZGIII, line 1: Consider adding a simple test in the likes on the one we have in the runtime for this; maybe we only need such a test in the pallet now.

crates/pallet-humanode-offences/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+use sp_staking::offence::{Offence, OffenceError, ReportOffence};
+
+/// The current storage version.
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+
+// We have to temporarily allow some clippy lints. Later on we'll send patches to substrate to
+// fix them at their end.
+#[allow(clippy::missing_docs_in_private_items)]
+#[frame_support::pallet]
+pub mod pallet {
```
MOZGIII, line 19: Previously, we had a convenient way to know how many offences happened on the chain over time. Let's add a simple counter of offences, so that we can still have that metric.

utils/checks/snapshots/features.yaml
```diff
@@ -2197,6 +2197,10 @@
 - name: pallet-grandpa 4.0.0-dev
   features:
     - std
+- name: pallet-humanode-offences 0.1.0
+  features:
+    - default
```
MOZGIII, line 2202: I wonder why `default` shows up here...

MOZGIII: Must be because it itself is a workspace target, so when it gets selected for tests it is included with `default` on.

dmitrylavrenov: In the same way as for other pallets in our code base.

MOZGIII: Yes; we might want to advance this further so it shows features per compilation target.

crates/pallet-humanode-offences/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1291/commits/e03ac3da01bb5cefc4049801c73d9e054f8b8a02

crates/pallet-humanode-offences/src/lib.rs
dmitrylavrenov: Implemented.

crates/humanode-runtime/src/lib.rs
```diff
@@ -863,7 +816,7 @@ construct_runtime!(
         TokenClaimsPot: pallet_pot::<Instance3> = 11,
         TransactionPayment: pallet_transaction_payment = 12,
         Session: pallet_session = 13,
-        Offences: pallet_offences = 14,
+        HumanodeOffences: pallet_humanode_offences = 14,
```
MOZGIII, line 819: Reusing the pallet bumber might cause issues, use a different number

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1291/commits/f970db1a7e547e20d015f4be55b3b11d865b3347

MOZGIII (approved):
Looks ready to me, except for the last issue with reusing the pallet number

crates/pallet-humanode-offences/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+            <Total<T>>::mutate(|total| {
+                let new_total = total.map_or(offences_number, |t| {
+                    t.checked_add(offences_number)
+                        .expect("u64 is big enough for this overflow to be practically impossible")
+                });
+                *total = Some(new_total);
+            });
+        }
+
+        Ok(())
```
quasiyoke, line 88: `pallet_offences / report_offence` deposits `Event::offence` — it's useless for us 'cause we have `pallet_bioauth / Event::AuthRemoved` ? I'd like to note that it's slightly less informative. This comment is just a reminder to discuss events and expected reasons to emit them.

crates/pallet-humanode-offences/src/mock.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+impl TestExternalitiesExt for frame_support::sp_io::TestExternalities {
+    fn execute_with_ext<R, E>(&mut self, execute: E) -> R
+    where
+        E: for<'e> FnOnce(&'e ()) -> R,
+    {
+        let guard = runtime_lock();
+        let result = self.execute_with(|| execute(&guard));
+        drop(guard);
```
quasiyoke, line 309:
```suggestion
        let result = {
            let guard = runtime_lock();
            self.execute_with(|| execute(&guard))
        };
```

Not insisting, just to discuss idioms.

quasiyoke (approved):

crates/pallet-humanode-offences/src/mock.rs
dmitrylavrenov: At the current moment the same style is used as for the rest our pallets,I guess it can be done in separate PR for all our pallets in our codebase if it's needed.

crates/pallet-humanode-offences/src/lib.rs
dmitrylavrenov: Good thoughts, I was thinking about it as well. But forgot to implement it. Thanks!

The general logic can be extended in theory (not just deauthentication) and it's gonna be good to explicitly submit events to the chain about offences if I would like to subscribe just on offence logic for example.

crates/pallet-humanode-offences/src/mock.rs
dmitrylavrenov: I like the idea of starting organizing a library of test/mock utilities. Will create an issue as a follow up of this conversation.

dmitrylavrenov: https://github.com/humanode-network/humanode/issues/1337

crates/pallet-humanode-offences/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1291/commits/6ef1117c2a975cbd370ab29048881c9cbb3d705b

MOZGIII (approved):

PR #1315, opened 2024-11-01
Rework migration logic

dmitrylavrenov:


crates/humanode-runtime/src/init_storage_version.rs
```diff
@@ -0,0 +1,62 @@
...
+        let onchain_storage_version = P::on_chain_storage_version();
+
+        let mut weight = R::DbWeight::get().reads(1);
+
+        if onchain_storage_version == 0 && current_storage_version != 0 {
+            info!(
+                "{}: Initializing onchain storage version to {:?}",
+                P::name(),
+                current_storage_version,
+            );
```
MOZGIII, line 34:
```suggestion
            info!(
                "{}: Initializing an unset on-chain storage version to {:?}, assuming the effective state version is the latest pallet version",
                P::name(),
                current_storage_version,
            );
```

MOZGIII (changes requested):
Please explain the logic behind this change - in particular why alter all the pallets

crates/pallet-balanced-currency-swap-bridges-initializer/src/tests.rs
```diff
@@ -411,7 +411,7 @@ fn runtime_upgrade() {
             assert_eq!(<LastForceRebalanceAskCounter<v1::Test>>::get(), 0);
 
             // Do runtime upgrade hook.
-            v1::AllPalletsWithoutSystem::on_runtime_upgrade();
+            UpgradeInit::<v1::Test>::on_runtime_upgrade();
```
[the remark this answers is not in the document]
dmitrylavrenov: The migration is not executed for each runtime upgrade logic in this rework. It should be explicitly included into runtime executive. 

So, we apply just a needed migration

crates/pallet-balanced-currency-swap-bridges-initializer/src/tests.rs
```diff
@@ -411,7 +411,7 @@ fn runtime_upgrade() {
             assert_eq!(<LastForceRebalanceAskCounter<v1::Test>>::get(), 0);
 
             // Do runtime upgrade hook.
-            v1::AllPalletsWithoutSystem::on_runtime_upgrade();
+            <(UpgradeInit<v1::Test>, v1::AllPalletsWithoutSystem)>::on_runtime_upgrade();
```
MOZGIII, line 414: Why not the `execute_update_init()`?

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
```diff
@@ -9,76 +13,54 @@ use crate::{
...
 
-/// Initialize the bridges pot accounts if required.
-pub fn on_runtime_upgrade<T: Config>() -> Weight {
-    let last_initializer_version = <LastInitializerVersion<T>>::get();
-    let last_force_rebalance_ask_counter = <LastForceRebalanceAskCounter<T>>::get();
-    let current_force_rebalance_ask_counter = T::ForceRebalanceAskCounter::get();
-
-    let mut weight = T::DbWeight::get().reads(2);
+/// Execute upgrade init.
+pub struct UpgradeInit<T>(sp_std::marker::PhantomData<T>);
```
MOZGIII, line 17: Why change this from migration to an explicit upgrade? Doesn't make sense to me actually...

MOZGIII:
> The main goal of this PR is to avoid running runtime upgrade execution for each pallet on each runtime upgrade. The idea is taken from `substrate` and `polkadot-sdk` approach to use an ability passing actual migration at the runtime itself.

Can you elaborate on this? There are migrations of the pallets that apply to the pallet itself more or less universally, and then there are circumstantial upgrades that may or may not apply to a chain.

Did substrate / polkadot-sdk change their ways by making all migrations explicit now? Otherwise I don't see how the changes you've made to all the pallets are justified. It makes sense for *some* things to be implemented on the runtime level - like a corrective action that is non-pallet-specific, i.e. fixing the storage after the global architecture design flaws are found and corrected - but not for all the pallets.

Did you see anything in the substrate that would support moving from the pallet hooks to the ad-hoc migration types en-masse?

dmitrylavrenov:
> Can you elaborate on this? There are migrations of the pallets that apply to the pallet itself more or less universally, and then there are circumstantial upgrades that may or may not apply to a chain.

Yep, there are separated objects at pallets that responsible for migration logic or `upgrade_init`- like logic we introduced to run force execution in some cases.

> Did substrate / polkadot-sdk change their ways by making all migrations explicit now? Otherwise I don't see how the changes you've made to all the pallets are justified. It makes sense for _some_ things to be implemented on the runtime level - like a corrective action that is non-pallet-specific, i.e. fixing the storage after the global architecture design flaws are found and corrected - but not for all the pallets.

During my investigation and research on migration logic at the polkadot itself I've found out that they used this approach all time. All migrations are introduced and run at runtime level when needed. If the migration is not used, then it was removed. There is an example - https://github.com/paritytech/polkadot/pull/7495.

> Did you see anything in the substrate that would support moving from the pallet hooks to the ad-hoc migration types en-masse?

There are a lot of PRs like mentioned above.


Additional reason of doing rework to have the same design and be able to apply migrations when they are introduced at substrate code base. Like the following one - https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/multisig/src/migrations.rs#L43. During substrate update to `polkadot-v0.9.43` work it was discovered that we need to run `pallet_multisig` migration that wasn't run before. So, we need to include it in our runtime. That's why I suggest to redesign our logic of executing  runtime upgrade logic.

crates/pallet-balanced-currency-swap-bridges-initializer/src/upgrade_init.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1315#issuecomment-2474717145

MOZGIII:
What version of substrate was this introduced at? If it is something from the future we shouldn't bother just yet.

Either way - let's separate the migration reworks and state version fix at runtime upgrade into separate PRs. One of them is uncontroversial and we'll be able to merge it quickly.

dmitrylavrenov:
> What version of substrate was this introduced at? If it is something from the future we shouldn't bother just yet.
> 
> Either way - let's separate the migration reworks and state version fix at runtime upgrade into separate PRs. One of them is uncontroversial and we'll be able to merge it quickly.

The approach is already in our forks code base. `pallet_balances` has it since 2020 and the actual migration as well.

crates/pallet-balanced-currency-swap-bridges-initializer/src/tests.rs
dmitrylavrenov: Reimplemented with `execute_on_runtime_upgrade`

MOZGIII:
We'd have to put this on hold for now, a I need to do more research on this and dig deeper to really conclude something propely; note there are no useless migrations, and, in principle, migrations are to be kept forever; unless there is a special certain cutoff date for them.

dmitrylavrenov:
> We'd have to put this on hold for now, a I need to do more research on this and dig deeper to really conclude something propely; note there are no useless migrations, and, in principle, migrations are to be kept forever; unless there is a special certain cutoff date for them.

Got it. The PR is not about avoiding having useless migrations, but to make the migrations itself more flexible. We can always have it at runtime level to keep it forever but control all migrations in one place.

Btw, looking forward to doing more research on your side and discussing later. I just was inspired by the general logic of custom upgrades usage after researching substrate updates related to `polkadot-v0.9.43`.

crates/pallet-dummy-precompiles-code/src/tests.rs
```diff
@@ -46,8 +44,8 @@ fn runtime_upgrade() {
                 0
             );
 
-            // Do runtime upgrade hook.
-            v1::AllPalletsWithoutSystem::on_runtime_upgrade();
+            // // Do runtime upgrade hook.
```
MOZGIII, line 47:
```suggestion
            // Do runtime upgrade hook.
```

crates/pallet-dummy-precompiles-code/src/mock/v1.rs
```diff
@@ -15,10 +15,11 @@ use frame_support::{
...
 
 use super::*;
-use crate::{self as pallet_dummy_precompiles_code};
+use crate::{self as pallet_dummy_precompiles_code, UpgradeInit};
 
-type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
-type Block = frame_system::mocking::MockBlock<Test>;
+type UncheckedExtrinsic =
+    frame_support::sp_runtime::testing::TestXt<RuntimeCall, frame_system::CheckEra<Test>>;
+type Block = frame_support::sp_runtime::testing::Block<UncheckedExtrinsic>;
```
MOZGIII, line 22: Interesting, why did we have to switch from one to the other?

MOZGIII: Because `Executive` needs it I deduce...

dmitrylavrenov:
@MOZGIII the idea is still relevant ? Or we would like to leave migration logic as it is 

In case, it's relevant, I'll rebase over latest master

MOZGIII:
This should be a part of one of the substrate upgrades, but it is not explicitly marked as such, and doesn't contain the necessary context for when a similar change in the thought was introduced to substrate.



> the idea is still relevant ? Or we would like to leave migration logic at our pallets as it is

This is, really, very simple. We want to follow the lead of the upstream here. So all that's necessary to decide is the context from above. Please add it to the PR description, and it will be quite trivial to make a decision.

Most likely we'll go with the upgrade. If it's not yet what the substrate was doing - we'll postpone it. If this is something that had to be included with our current substrate version - let's apply it.

MOZGIII (approved):
The changes here are technically fine - but there is still a question of whether we want to switch to this approach or not.

See https://github.com/humanode-network/humanode/pull/1315#issuecomment-2817363941

PR #1325, opened 2024-11-07
Add `with-tracing` feature at `humanode-runtime` and enable it for `try-runtime` execution

dmitrylavrenov:
Analysed substrate repo for tracing support. `with-tracing` feature is defined and used at:
- `frame-executive`
- `sp-io`
- `sp-tracing`

Also, found the following guide - https://github.com/humanode-network/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/rpc-api/src/state/mod.rs#L157

Tracing support is added in the same way as `frame-executive` has - https://github.com/humanode-network/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/executive/src/lib.rs#L417.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -78,7 +78,7 @@ jobs:
           - suite: "base"
             build-args: ""
           - suite: "try-runtime"
-            build-args: "--features try-runtime,with-tracing"
+            build-args: "--features try-runtime"
       fail-fast: false
     name: End-to-end tests / ${{ matrix.test.suite }}
     runs-on: ubuntu-22.04
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -78,7 +78,7 @@ jobs:
           - suite: "base"
             build-args: ""
           - suite: "try-runtime"
-            build-args: "--features try-runtime"
+            build-args: "--features try-runtime,with-tracing"
       fail-fast: false
     name: End-to-end tests / ${{ matrix.test.suite }}
     runs-on: ubuntu-22.04
--- a/crates/humanode-runtime/Cargo.toml
+++ b/crates/humanode-runtime/Cargo.toml
@@ -94,7 +94,6 @@ sp-runtime = { workspace = true }
 sp-session = { workspace = true }
 sp-staking = { workspace = true }
 sp-std = { workspace = true }
-sp-tracing = { workspace = true }
 sp-transaction-pool = { workspace = true }
 sp-version = { workspace = true }
 static_assertions = { workspace = true }
@@ -224,7 +223,6 @@ std = [
   "sp-session/std",
   "sp-staking/std",
   "sp-std/std",
-  "sp-tracing/std",
   "sp-transaction-pool/std",
   "sp-version/std",
   "substrate-wasm-builder",
@@ -274,5 +272,4 @@ try-runtime = [
 with-tracing = [
   "frame-executive/with-tracing",
   "sp-io/with-tracing",
-  "sp-tracing/with-tracing"
 ]
--- a/crates/humanode-runtime/Cargo.toml
+++ b/crates/humanode-runtime/Cargo.toml
@@ -88,6 +88,7 @@ sp-block-builder = { workspace = true }
 sp-consensus-babe = { workspace = true }
 sp-core = { workspace = true }
 sp-inherents = { workspace = true }
+sp-io = { workspace = true }
 sp-offchain = { workspace = true }
 sp-runtime = { workspace = true }
 sp-session = { workspace = true }
@@ -106,7 +107,6 @@ ethereum = { workspace = true, features = ["default"] }
 hex = { workspace = true }
 hex-literal = { workspace = true }
 serde_json = { workspace = true }
-sp-io = { workspace = true }
 sp-keystore = { workspace = true }
 
 [features]
@@ -270,4 +270,9 @@ try-runtime = [
   "pallet-vesting/try-runtime",
   "primitives-currency-swap-proxy/try-runtime",
   "sp-runtime/try-runtime",
+]
+with-tracing = [
+  "frame-executive/with-tracing",
+  "sp-io/with-tracing",
+  "sp-tracing/with-tracing"
 ]
--- a/crates/humanode-runtime/src/lib.rs
+++ b/crates/humanode-runtime/src/lib.rs
@@ -1584,6 +1584,8 @@ impl_runtime_apis! {
     #[cfg(feature = "try-runtime")]
     impl frame_try_runtime::TryRuntime<Block> for Runtime {
         fn on_runtime_upgrade(checks: frame_try_runtime::UpgradeCheckSelect) -> (Weight, Weight) {
+            sp_io::init_tracing();
+
             // NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
             // have a backtrace here. If any of the pre/post migration checks fail, we shall stop
             // right here and right now.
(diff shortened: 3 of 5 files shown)


crates/humanode-runtime/src/lib.rs
```diff
@@ -1597,14 +1599,8 @@ impl_runtime_apis! {
...
             select: frame_try_runtime::TryStateSelect
         ) -> Weight {
-            sp_tracing::info!(
-                target: "humanode-runtime",
-                "try-runtime: executing block {:?} / root checks: {:?} / signature check: {:?} / try-state-select: {:?}",
-                block.header.hash(),
-                state_root_check,
-                signature_check,
-                select,
-            );
```
dmitrylavrenov, line 1607: Already have it under `try_execute_block` - https://github.com/humanode-network/substrate/blob/56b16fefbd7a5ba07afd94cb37367d078da0ac52/frame/executive/src/lib.rs#L245

quasiyoke (approved):
It's a pity that they call this feature `with-tracing`; as @MOZGIII mentioned, features are additive and the `with-` prefix is ​​redundant. However, I think it's more important to do it "like they do", uniformly, than to do it idiomatically.

crates/humanode-runtime/Cargo.toml
```diff
@@ -271,3 +269,7 @@ try-runtime = [
   "primitives-currency-swap-proxy/try-runtime",
   "sp-runtime/try-runtime",
 ]
+with-tracing = [
```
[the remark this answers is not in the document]
dmitrylavrenov: Substrate doesn't provide it by default, mentioned it in the PR description - https://github.com/humanode-network/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/rpc-api/src/state/mod.rs#L157

crates/humanode-runtime/src/lib.rs
```diff
@@ -1584,6 +1584,8 @@ impl_runtime_apis! {
     #[cfg(feature = "try-runtime")]
     impl frame_try_runtime::TryRuntime<Block> for Runtime {
         fn on_runtime_upgrade(checks: frame_try_runtime::UpgradeCheckSelect) -> (Weight, Weight) {
+            sp_io::init_tracing();
```
MOZGIII, line 1587: Why do we need to do manual initialization? I kind of assumed this would be handled by the framework, such that we don't have to; it also seems like this is not the place to do it - of course, unless it is justified by some logic was not communicated anywhere here

dmitrylavrenov: It's not handled by framework now, tested. As I got, it's just developed to trace just transactions execution by enabling was tracing manually. The same as here - https://github.com/humanode-network/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/rpc-api/src/state/mod.rs#L157

dmitrylavrenov: It's enabled just for some parts of code, not everywhere.

.github/workflows/code.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           - suite: "base"
             build-args: ""
           - suite: "try-runtime"
-            build-args: "--features try-runtime"
+            build-args: "--features try-runtime,with-tracing"
```
MOZGIII, line 81: I don't like this - enabling something like this via CI would result in a narrow case of when this feature is on. I think we shouldn't enable it just for CI - like if you build and run node for try-runtime locally you'd expect it to work the same way as CI.

So, at the very least, I'd suggest providing an easier route flag for the end user to enable this feature - if we need it.

crates/humanode-runtime/src/lib.rs
MOZGIII: Should we correct this at upstream?

crates/humanode-runtime/Cargo.toml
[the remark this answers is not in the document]
MOZGIII: Please show me where, I must be missing somethig

crates/humanode-runtime/src/lib.rs
dmitrylavrenov: It's a good question, but it can take a lot of time on agreeing where it should be enabled or not. As I got, their goal  was just tracing transaction logic execution.

crates/humanode-runtime/Cargo.toml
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/executive/Cargo.toml#L41

.github/workflows/code.yml
dmitrylavrenov: Got your point, removed from CI - https://github.com/humanode-network/humanode/pull/1325/commits/1e92f329135f06392a60021a0c5640936669c91e

crates/humanode-runtime/Cargo.toml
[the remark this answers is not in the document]
MOZGIII: Reading up on the comments in that PR - it seems like `tracing` is not used for logging in the general purpose, but for something else that is kind of very specific and that I didn't anticipate they'd design like that. In particular - there is a specific feature for block re-execution with tracing. Who would want that and why is a mystery to me as of now - but at least it is now clear there is a distinction in the purpose of the regular logging infrastructure and `sp_tracing` logging. It is clear that the difference is beyond them just being two logging frontends - as they seem to be targeting different backends - so we have to learn and figure out when to use which.

To be more explicit here - I clearly was wrong about always using `tracing`. I think that, by default, we'd actually want to use `frame_support::log` APIs.

We also must figure out what they use `tracing` for.

crates/humanode-runtime/src/lib.rs
MOZGIII: What's odd is that they use `frame_support::log`; does this mean we got this message kind of printed twice with tracing enabled?

dmitrylavrenov: Yep, it's printed twice
image

Without - just once, right.

crates/humanode-runtime/src/lib.rs
MOZGIII: Looks like we need to do some more research here. It might be that they messed this up - or the while tracing integration is a not fully thought through thing. If they only need it for that tracing transaction logic execution and they don't enable it for `try_runtime` - why should we?

dmitrylavrenov: Your explanation above - https://github.com/humanode-network/humanode/pull/1325#discussion_r1832709696 sounds right, and I guess we shouldn't enable it. It's enough just use `frame_support::log` for logging `on_runtime_upgrade` execution to log operations if we need it.

crates/humanode-runtime/Cargo.toml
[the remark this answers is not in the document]
dmitrylavrenov: Sounds reasonable, I agree on using `frame_support::log` APIs for just logging goals in our codebase. 

From my understanding, they use `tracing` to connect traced execution of different codes parts using spans to have the whole flow.

As an example, from entrypoint of `init_block` to error happened inside some pallet.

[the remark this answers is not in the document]
dmitrylavrenov: As for me, using log APIs looks good as well - https://github.com/humanode-network/humanode/pull/1326.

We don't need to enable any features here to have these logs.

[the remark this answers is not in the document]
MOZGIII: The question is when would you need to use tracing (and when not - and use just `log`). Until this question resolved we don't have an understanding of this domain. This should be addressed, can you create an issue?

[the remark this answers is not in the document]
dmitrylavrenov: Agree, it's good domain to be addressed on our side, create - https://github.com/humanode-network/humanode/issues/1327

crates/humanode-runtime/src/lib.rs
MOZGIII: Seems like adding the feature to the runtime to enable the nested futures is the only takeaway here - cause that's just a strait-up improvement; but let's add it in a separate PR, and close this one as is for historical preservation of the context.

PR #1326, opened 2024-11-07
Use `frame_support::logs` APIs

dmitrylavrenov:
As a follow up of https://github.com/humanode-network/humanode/pull/1325#discussion_r1832709696

### DIFF
--- a/crates/pallet-humanode-session/src/migrations/v1.rs
+++ b/crates/pallet-humanode-session/src/migrations/v1.rs
@@ -2,7 +2,7 @@
 
 use frame_support::pallet_prelude::*;
 use frame_support::storage_alias;
-use frame_support::{dispatch::GetStorageVersion, sp_tracing::info, traits::Get, weights::Weight};
+use frame_support::{dispatch::GetStorageVersion, log::info, traits::Get, weights::Weight};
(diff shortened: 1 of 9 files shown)


crates/pallet-humanode-session/src/migrations/v1.rs
```diff
@@ -27,10 +27,10 @@ pub fn migrate<T: Config>() -> Weight {
     // Read the onchain version.
     let mut weight: Weight = T::DbWeight::get().reads(1);
 
-    info!(message = "Running migration to v1", from = ?onchain);
+    info!("Running migration to v1");
```
MOZGIII, line 30:
```suggestion
    info!("Running migration to v1 from {onchain:?}");
```

MOZGIII (approved):
This looks trivial enough, and we can merge it as it would not be a big deal to change it back where these changes are unwarranted after we find out when to use what properly.
So, we can merge it.
Usually, I'd say we're jumping the gun with those changes - but currently, since the `with-tracing` features are not enabled in places, the logging points we left in the code just don't work, which seems to be an inferior state than before - so gonig forward with this now is justified.

quasiyoke (approved):

Issue #223, opened 2021-12-12
Get rid of compile-time features for switching the facetec integration tests (`logic-integration-tests`)


MOZGIII:
Switch to skipping the test if a certain env var isn't set instead.

MOZGIII:
@quasiyoke could you maybe take this one?

PR #1347, opened 2024-11-29
Enable FaceTec integration tests unconditional compilation; ignore them in test runtime

quasiyoke:
We want to ensure that the test build was checked in CI until there was a major reconstruction of bioauth flow and robonode.

```sh
$ cargo test
...
test logic::tests::double_enroll ... ignored, requires Facetec server
test http::tests::get_facetec_device_sdk_params_in_dev_mode ... ok
test logic::tests::enroll_authenticate ... ignored, requires Facetec server
test http::tests::enroll_error_person_already_enrolled ... ok
test logic::tests::first_authenticate ... ignored, requires Facetec server
test logic::tests::standalone_enroll ... ignored, requires Facetec server
...
$ # Doesn't require rebuild:
$ cargo test -- --include-ignored
...
test logic::tests::enroll_authenticate ... FAILED
test logic::tests::double_enroll ... FAILED
test logic::tests::first_authenticate ... FAILED
test logic::tests::standalone_enroll ... FAILED
...
```

✓ Made sure that high-level compilation errors in tests are checked (for example, type errors).

Considered other options to ignore tests after enabling their unconditional compilation:
- Controlling ignoring tests via env var — more granular than `#[ignore]`, but not in general use. Requires to add exit logic to each test. To avoid it, one could use a declarative macro, but I think there would be little benefit from it and macros should be used only as a last resort.
- Test harness Nextest, along with the `#[ignore]` attribute, [offers to set a "default filter"](https://nexte.st/docs/running/#running-a-subset-of-tests-by-default). For us, using Nextest would be an overkill for now. Nextest requires using a custom CLI testing command, which we'd like to avoid for now.

Closes: #223

### DIFF
--- a/crates/robonode-server/src/logic/tests.rs
+++ b/crates/robonode-server/src/logic/tests.rs
@@ -1,5 +1,3 @@
-#![cfg(feature = "logic-integration-tests")]
-
 use std::marker::PhantomData;
 
 use facetec_api_client as ft;
@@ -174,6 +172,7 @@ async fn setup() -> (
 }
 
 #[tokio::test]
+#[ignore = "requires FaceTec server"]
 #[tracing_test::traced_test]
 async fn standalone_enroll() {
     let (_guard, test_params, logic) = setup().await;
@@ -189,6 +188,7 @@ async fn standalone_enroll() {
 }
 
 #[tokio::test]
+#[ignore = "requires FaceTec server"]
 #[tracing_test::traced_test]
 async fn first_authenticate() {
     let (_guard, test_params, logic) = setup().await;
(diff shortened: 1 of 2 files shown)


crates/robonode-server/src/logic/tests.rs
```diff
@@ -174,6 +172,7 @@ async fn setup() -> (
 }
 
 #[tokio::test]
+#[ignore = "requires Facetec server"]
```
MOZGIII, line 175:
```suggestion
#[ignore = "requires FaceTec server"]
```

MOZGIII (approved):

crates/robonode-server/src/logic/tests.rs
quasiyoke: Done

dmitrylavrenov (approved):

PR #1362, opened 2024-12-11
Temporarily ignore RUSTSEC-2024-0421

dmitrylavrenov:
https://rustsec.org/advisories/RUSTSEC-2024-0421

We have internal dependencies (libp2p) that depends on vulnerable idna version.

Based on the vulnerability description looks like we can ignore it for now.


quasiyoke (approved):
The vulnerability would be significant for us if we allowed users to distribute URLs to other users and we had, say, a blacklist of some resources. AFAIK, this vulnerability should not cause problems for us.

MOZGIII:
This is actually an exploitable issue

MOZGIII (approved):

PR #1366, opened 2024-12-12
Refactor the CheckBioauthTx

MOZGIII:
This PR addresses the issue we've discovered during the code review with the too restrictive validation for the pallet bioauth extrinsics for the calls other than authenticate.

PR #1368, opened 2024-12-12
Correct the token claims validation logic CheckTokenClaim

MOZGIII:
Analogus to #1366, this is a slight correction to the `CheckTokenClaim`

### DIFF
--- a/crates/pallet-token-claims/src/signed_ext.rs
+++ b/crates/pallet-token-claims/src/signed_ext.rs
@@ -9,6 +9,7 @@ use frame_support::{
     traits::IsSubType,
     unsigned::{TransactionValidity, TransactionValidityError},
 };
+use primitives_ethereum::{EcdsaSignature, EthereumAddress};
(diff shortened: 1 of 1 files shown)


crates/pallet-token-claims/src/signed_ext.rs
```diff
@@ -98,7 +92,23 @@ where
...
+        }
+    }
+
+    fn validate_unsigned(
+        call: &Self::Call,
+        _info: &DispatchInfoOf<Self::Call>,
+        _len: usize,
+    ) -> TransactionValidity {
+        match call.is_sub_type() {
+            // Disallow `claim` call - can't be used an unsigned tx.
```
MOZGIII, line 110:
```suggestion
            // Disallow `claim` call - can't be used as an unsigned tx.
```

dmitrylavrenov (approved):
Like the explicit invalid call for unsigned case!

LGTM

quasiyoke (approved):

PR #1398, opened 2025-01-13
E2E: make sure that contract account's state doesn't vanish after emptying its wallet

quasiyoke:


.gitignore
```diff
@@ -10,3 +10,4 @@
 !/.yarn/releases
 !/.yarn/sdks
 !/.yarn/versions
+/utils/e2e-tests/ts/node_modules
```
MOZGIII, line 13: Should not be required, as we are using a subgle global `node_modules` dir at the root level...

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
```diff
@@ -0,0 +1,78 @@
...
+    await build(factoryAddress, bob, publicClient);
+
+    // Contract factory's `CREATE` nonce for the 3rd `build` should be 3.
+    // In the buggy EVM nonce = 1, the same as for the 1st `build`; transaction will be reverted.
+    const item3Address = await build(factoryAddress, bob, publicClient);
+    expect(item1Address).not.toBe(item3Address);
+  });
+});
+
+async function build(
```
MOZGIII, line 60: Why do we need this fn?

MOZGIII: As in, why not just send a tx, why all this?

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
```diff
@@ -0,0 +1,78 @@
...
+      abi: contractsFactory.abi,
+      bytecode: contractsFactory.bytecode,
+      value: 1n,
+    });
+    const deployContractTxReceipt =
+      await publicClient.waitForTransactionReceipt({
+        hash: deployContractTxHash,
+        timeout: 18_000,
+      });
+    const factoryAddress = deployContractTxReceipt.contractAddress!;
```
MOZGIII, line 34: Better use a real `assert` here instead of `!`. I can see how this can becode a problem if the code doesn't properly check this value.

utils/e2e-tests/ts/lib/abis/contractsFactory.ts
```diff
@@ -0,0 +1,66 @@
...
+//
+//     function withdrawAll() external isOwner {
+//         (bool success,) = owner.call{value:address(this).balance}("");
+//         require(success, "Transfer failed!");
+//     }
+//
+//     /**
+//      * Makes sure if caller is owner.
+//      */
+//     modifier isOwner() {
```
MOZGIII, line 28: Do we need this check? I think we should simplify the test contract, and this here seems like something that the trst can work fine without

MOZGIII (approved):
This PR has a lot of nits, and probably things that are worth discussion - but it looks like it should capture the bug precisely!

I suggest adding an assertion or a debug print of the nonce of the contract in between the calls to show it nore obviosuly where the fail is.

Good stuff!

.gitignore
quasiyoke: Solved this in a proper way at #1401

utils/e2e-tests/ts/lib/abis/contractsFactory.ts
quasiyoke: I agree, simplified

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
quasiyoke: ✅

utils/e2e-tests/ts/vitest.config.ts
```diff
@@ -18,6 +18,6 @@ export default defineConfig({
         singleThread: true,
       },
     },
-    testTimeout: 30_000,
+    testTimeout: 40_000,
```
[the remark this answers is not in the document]
quasiyoke: ✅

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    // Contract factory's `CREATE` nonce for the 3rd `build` should be 3.
+    // In the buggy EVM execution: nonce = 1, the same as for the 1st `build`; transaction will be reverted.
+    const build3Tx = await alice.writeContract(buildParams);
+    const build3Receipt = await publicClient.waitForTransactionReceipt({
+      hash: build3Tx,
+      timeout: 18_000,
+    });
+    expect(build3Receipt.status, "status of third `build`").toBe("success");
+  }, 35_000);
```
quasiyoke, line 88: This is a timeout. Unfortunately, they deprecated the API that allowed writing:
```suggestion
  }, { timeout: 35_000 });
```
I still hope that this test can be executed within 30s. On my machine, it takes 28s.

utils/e2e-tests/ts/lib/abis/contractsFactory.ts
```diff
@@ -0,0 +1,54 @@
...
+//         return new Item();
+//     }
+//
+//     // This method must be called by the contract owner. We do not check this invariant
+//     // directly for the sake of simplicity of the test.
+//     function withdrawAll() external {
+//         address owner = msg.sender; // `msg.sender` is sender of current call = contract deployer
+//         (bool success,) = owner.call{value:address(this).balance}("");
+//         require(success, "Transfer failed!");
+//     }
```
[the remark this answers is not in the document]
quasiyoke: Thanks! Removed this fluff

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
```diff
@@ -0,0 +1,89 @@
...
+    // Contract factory's `CREATE` nonce for the 2nd `build` should be 2 (in the buggy EVM execution: 0).
+    const build2Promise = alice.writeContract(buildParams).then((txHash) =>
+      publicClient.waitForTransactionReceipt({
+        hash: txHash,
+        timeout: 18_000,
+      }),
+    );
+    const [build1Receipt, withdrawalReceipt, build2Receipt] = await Promise.all(
+      [build1Promise, withdrawalPromise, build2Promise],
+    );
```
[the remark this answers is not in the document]
quasiyoke: These calls are no longer needed

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
quasiyoke: There are much fewer calls and the test duration has decreased, it takes about 24s on my machine now

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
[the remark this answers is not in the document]
quasiyoke: > should be more specifically named after the bug

We've chosen `tests/eth/contractsState.ts`: "contract account's state > doesn't vanish after emptying its wallet"

> tracking issue

Dima'll compose it soon

utils/e2e-tests/ts/tests/eth/contractsState.ts
```diff
@@ -0,0 +1,63 @@
...
+        timeout: 18_000,
+      });
+    expect(deployContractTxReceipt.status).toBe("success");
+    const contract = deployContractTxReceipt.contractAddress;
+    assert(contract);
+
+    // The nonce of the contract account immediately after creation.
+    const INITIAL_NONCE = 1;
+
+    // If there's a bug in the EVM execution, it will clear the contract account state (including nonce) after `withdrawAll`.
```
dmitrylavrenov, line 43: I would say not from point of view that if there's a bug, but the goal we do it and what we want to verify by this call

utils/e2e-tests/ts/tests/eth/contractsState.ts
```diff
@@ -0,0 +1,63 @@
...
+      gas: 30_585n,
+    });
+    const withdrawalReceipt = await publicClient.waitForTransactionReceipt({
+      hash: withdrawalTx,
+      timeout: 18_000,
+    });
+    expect(withdrawalReceipt.status, "status of withdrawal").toBe("success");
+
+    // Ethereum RPC returns the account nonce in the `eth_getTransactionCount` RPC call.
+    // https://github.com/humanode-network/frontier/blob/locked/polkadot-v0.9.42-2024-09-12/client/rpc/src/eth/state.rs#L116-L169
```
dmitrylavrenov, line 57: What about using the link to frontier master code ?

dmitrylavrenov (approved):
LGTM!

utils/e2e-tests/ts/tests/eth/contractsFactory.ts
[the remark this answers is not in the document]
quasiyoke: > huge comment explaining the bug itself

Perhaps a link to the tracking issue will provide the necessary context?

utils/e2e-tests/ts/tests/eth/contractsState.ts
quasiyoke: ✅

utils/e2e-tests/ts/tests/eth/contractsState.ts
MOZGIII: Actually better to link to a specific commit - that way we can ref the exact code

utils/e2e-tests/ts/tests/eth/contractsState.ts
MOZGIII, line 1: `contractsState` is way too generic. Better call it `issue1402` or `noNonceResetOnZeroBalance`

quasiyoke: Called it `preserveNonceOnZeroContractBalance`

utils/e2e-tests/ts/tests/eth/contractsState.ts
quasiyoke: Specified https://github.com/humanode-network/frontier/blob/1afab28e8d5aebe7d44f9043b3ba19e9555123dc/client/rpc/src/eth/state.rs#L116-L169

MOZGIII: See https://eips.ethereum.org/EIPS/eip-7702
This is a bit more complicated since EOAs could have code, too, and EIP 7702 specifies one way of doing it.

Not saying this enum is no good already - didn't think through yet, but just giving you guys a pointer since I was this discussion.

PR #1401, opened 2025-01-15
Improve E2E development experience

quasiyoke:


utils/e2e-tests/ts/lib/node.ts
```diff
@@ -6,6 +6,7 @@ import { AddCleanup } from "./cleanup";
 
 export type RunNodeParams = {
   args: string[];
+  env?: NodeJS.ProcessEnv | undefined;
```
MOZGIII, line 9:
```suggestion
  env?: NodeJS.ProcessEnv;
```

`env?: ...` implies `undefined`

utils/e2e-tests/ts/vitest.config.ts
```diff
@@ -20,4 +20,5 @@ export default defineConfig({
     },
     testTimeout: 30_000,
   },
+  cacheDir: "../../../node_modules/.e2e-vite/", // Default value: "./node_modules/.vite/"
```
MOZGIII, line 23: Should be a separate PR?

MOZGIII (approved):

utils/e2e-tests/ts/vitest.config.ts
MOZGIII: Is this path relative to the working directory of the process, or tot he config file?

quasiyoke: > Should be a separate PR?

In my opinion this is also related to development experience improvement 🙃

> path relative to the working directory of the process, or to the config file?

[They call it](https://vite.dev/config/shared-options.html#cachedir) "relative to the project root" — in practice, the path is relative to the config file. We can also specify an absolute path. Commented on this.

MOZGIII: > In my opinion this is also related to development experience improvement 🙃

Yes, I only suggest splitting this cause this we discuss while the seconds half of this PR could've been merged already.

It also implies that these development experience improvements are independent and thus there is no need to group them in a single PR. Naming of the PR does not excuse it.

MOZGIII: It also implies that these development experience improvements are independent and thus there is no need to group them in a single PR. Naming of the PR does not excuse it.

MOZGIII (approved):

Issue #1402, opened 2025-01-16
Tracking issue for reverted transaction that should succeed on EVM 2024-12


dmitrylavrenov:
This is a tracking issue for investigating the reason of reverted transaction that should succeed on EVM, getting root cause, reproducing the issue, implementing a proper fix.

---

Bugfix PRs:

- https://github.com/humanode-network/frontier/pull/147
- https://github.com/humanode-network/frontier/pull/140

Internal playground utils:
- https://github.com/humanode-network/humanode-playground/pull/70

dmitrylavrenov:
`succeded` transaction:
- trx: https://humanode.subscan.io/tx/0x722c86b6daa736487b5ade80d659cc91cdfcf62f3809a25f0ccf90726e341d9b
- exported state at https://humanode.subscan.io/block/7927643
- patched state to be able run dev mode with sudo non-zero alice
- run `humanode-peer` on 7927643 state => resubmitted transaction => `success`
- run `humanode-peer` on 7927643 state => run runtime upgrade to `v0.0.0-it6-2024-10-17` (current mainnet) => resubmitted transaction => `success`
- run `humanode-peer` on 7927643 state => run runtime upgrade to current draft `polkadot-v0.9.43`  => resubmitted transaction => `success`

`failed` transaction:
- trx: https://humanode.subscan.io/tx/0x332e1eb3400fb9654775f5916ce2cdb74a9c309e2ffd99009df558672c47ab04
- exported state at https://humanode.subscan.io/block/10507907
- patched state to be able run dev mode with sudo non-zero alice
- run `humanode-peer` on 10507907 state => resubmitted transaction => `reverted`
- run `humanode-peer` on 10507907 state => run runtime upgrade to current draft `polkadot-v0.9.43`  => resubmitted transaction => `reverted`

dmitrylavrenov:
- prepared the `humanode-peer` version with enabled debug of evm opcodes operation taking into account the value of left gas
- rerun the same process as above for succeeded (block number 7927643) and failed (block number 10507907) transactions

[logs-7927643.txt](https://github.com/user-attachments/files/18442478/logs-7927643.txt)
[logs-10507907.txt](https://github.com/user-attachments/files/18442477/logs-10507907.txt)

dmitrylavrenov:
- added more debug logs
- discovered the root cause why is the failed transaction reverted

[...]

PR #1433, opened 2025-02-10
Migration recovering broken EVM contract nonces

quasiyoke:
Migration for  #1402 

Run on try-runtime:

```
2025-02-13 17:31:39.279  INFO main frame_support::migrations: Found Offences keys pre-removal 👀
2025-02-13 17:31:40.218  INFO main frame_support::migrations: Removed 80814 Offences keys 🧹
2025-02-13 17:31:45.177  INFO main frame_support::migrations: No Offences keys found post-removal 🎉
2025-02-13 17:31:45.178  INFO main evm_nonces_recovery: EvmSystem: Running migration to recover broken nonces
2025-02-13 17:31:45.187  INFO main evm_nonces_recovery: Account 0x38d7…0e44 has zero nonce
2025-02-13 17:31:45.187  INFO main evm_nonces_recovery: Account 0x38d7…0e44 minimal valid nonce is 1
2025-02-13 17:31:45.187  INFO main evm_nonces_recovery: Account 0xae8d…1cd6 lacks its state
2025-02-13 17:31:45.188  INFO main evm_nonces_recovery: Account 0xae8d…1cd6 minimal valid nonce is 1
2025-02-13 17:31:45.192  INFO main evm_nonces_recovery: Account 0xdc8a…0414 lacks its state
2025-02-13 17:31:45.192  INFO main evm_nonces_recovery: Account 0xdc8a…0414 minimal valid nonce is 1
2025-02-13 17:31:45.194  INFO main evm_nonces_recovery: Account 0x45e7…51b3 lacks its state
2025-02-13 17:31:45.194  INFO main evm_nonces_recovery: Account 0x45e7…51b3 minimal valid nonce is 2
2025-02-13 17:31:45.198  INFO main evm_nonces_recovery: Account 0x278e…29c4 lacks its state
2025-02-13 17:31:45.198  INFO main evm_nonces_recovery: Account 0x278e…29c4 minimal valid nonce is 1
2025-02-13 17:31:45.198  INFO main evm_nonces_recovery: Account 0x8991…ae8e lacks its state
2025-02-13 17:31:45.199  INFO main evm_nonces_recovery: Account 0x8991…ae8e minimal valid nonce is 1
2025-02-13 17:31:45.201  INFO main evm_nonces_recovery: Account 0xbcf7…a775 lacks its state
2025-02-13 17:31:45.201  INFO main evm_nonces_recovery: Account 0xbcf7…a775 minimal valid nonce is 1
2025-02-13 17:31:45.204  INFO main evm_nonces_recovery: Account 0x7b7a…2dce has zero nonce
2025-02-13 17:31:45.205  INFO main evm_nonces_recovery: Account 0x7b7a…2dce minimal valid nonce is 1
2025-02-13 17:31:45.205  INFO main evm_nonces_recovery: Account 0xf0b9…7c72 lacks its state
2025-02-13 17:31:45.205  INFO main evm_nonces_recovery: Account 0xf0b9…7c72 minimal valid nonce is 2
2025-02-13 17:31:45.205  INFO main evm_nonces_recovery: Account 0x9060…7677 lacks its state
2025-02-13 17:31:45.205  INFO main evm_nonces_recovery: Account 0x9060…7677 minimal valid nonce is 1
2025-02-13 17:31:45.206  INFO main evm_nonces_recovery: Account 0xb843…7978 lacks its state
2025-02-13 17:31:45.206  INFO main evm_nonces_recovery: Account 0xb843…7978 minimal valid nonce is 1
2025-02-13 17:31:45.207  INFO main evm_nonces_recovery: Account 0xc796…51c5 lacks its state
2025-02-13 17:31:45.208  INFO main evm_nonces_recovery: Account 0xc796…51c5 minimal valid nonce is 1
2025-02-13 17:31:45.208  INFO main evm_nonces_recovery: Account 0x8c20…5441 lacks its state
2025-02-13 17:31:45.208  INFO main evm_nonces_recovery: Account 0x8c20…5441 minimal valid nonce is 1
2025-02-13 17:31:45.209  INFO main evm_nonces_recovery: Account 0x820f…c02a lacks its state
2025-02-13 17:31:45.209  INFO main evm_nonces_recovery: Account 0x820f…c02a minimal valid nonce is 1
2025-02-13 17:31:45.209  INFO main evm_nonces_recovery: Account 0x6de8…bcce lacks its state
2025-02-13 17:31:45.209  INFO main evm_nonces_recovery: Account 0x6de8…bcce minimal valid nonce is 1
2025-02-13 17:31:45.210  INFO main evm_nonces_recovery: Account 0x30f1…eebc lacks its state
2025-02-13 17:31:45.211  INFO main evm_nonces_recovery: Account 0x30f1…eebc minimal valid nonce is 1
2025-02-13 17:31:45.212  INFO main evm_nonces_recovery: Account 0x4899…c568 has zero nonce
2025-02-13 17:31:45.212  INFO main evm_nonces_recovery: Account 0x4899…c568 minimal valid nonce is 1
2025-02-13 17:31:45.215  INFO main evm_nonces_recovery: Account 0xd64c…701e has zero nonce
2025-02-13 17:31:45.215  INFO main evm_nonces_recovery: Account 0xd64c…701e minimal valid nonce is 1
2025-02-13 17:31:45.218  INFO main evm_nonces_recovery: Account 0xecd6…c645 lacks its state
2025-02-13 17:31:45.219  INFO main evm_nonces_recovery: Account 0xecd6…c645 minimal valid nonce is 1
2025-02-13 17:31:45.222  INFO main evm_nonces_recovery: EvmSystem: Migrated
...
2025-02-13 17:31:52.509  INFO main try-runtime::cli: TryRuntime_on_runtime_upgrade executed without errors. Consumed weight = (10134000000000 ps, 0 byte), total weight = (2000000000000 ps, 18446744073709551615 byte) (506.70 %, 0.00 %).
Test succeeded
```


crates/pallet-evm-system/src/migrations/recover_broken_nonces.rs
```diff
@@ -0,0 +1,187 @@
...
+///
+/// [1]: https://ethereum.github.io/yellowpaper/paper.pdf#section.7
+fn contract_address(sender: &H160, nonce: impl rlp::Encodable) -> H160 {
+    let mut rlp = RlpStream::new_list(2);
+    rlp.append(sender);
+    rlp.append(&nonce);
+    /// Address is the rightmost 160 bits of hash.
+    const ADDR_OFFSET: usize = (256 - 160) / 8;
+    H160::from_slice(&keccak_256(&rlp.out())[ADDR_OFFSET..])
+}
```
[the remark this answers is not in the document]
MOZGIII: `nonce: impl rlp::Encodable` - seems too broad; how about making this type requirement stricter?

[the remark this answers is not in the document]
dmitrylavrenov: What about using `alloy-primitives` - https://crates.io/crates/alloy-primitives? Is it good option?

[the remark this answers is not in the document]
MOZGIII: Wouldn't want to introduce new deps here for this

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+        info!("Account {id} minimal valid nonce is {nonce:?}");
+        (nonce, weight)
+    }
+}
+
+/// Contract address that will be produced by the [`CREATE` opcode][1].
+///
+/// [1]: https://ethereum.github.io/yellowpaper/paper.pdf#section.7
+fn contract_address(sender: &H160, nonce: impl rlp::Encodable) -> H160 {
```
dmitrylavrenov, line 163: In case we use own implementation, what about moving this to `crypto-utils` or `crypto-utils-evm`? It would be helpful to reuse it later.

MOZGIII: But it would kind of is also be fine to leave here - in a sense, this is a highly specialized conversion for this migration; we wouldn't want it to be updated/changed if the shared implementation changes.

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
```diff
@@ -0,0 +1,189 @@
...
+use crate::{Account, AccountInfo, Config, Pallet};
+
+/// EVM state provider.
+pub trait EvmStateProvider<AccountId> {
+    /// Check whether account address has code.
+    fn has_code(account_id: &AccountId) -> (bool, Weight);
+
+    /// Gives an iterator over accounts managed by EVM. Considers precompiled contracts' accounts
+    /// as not "managed by EVM".
+    fn accounts_managed_by_evm() -> impl Iterator<Item = (AccountId, Weight)>;
```
dmitrylavrenov, line 20: Just to think, based on conversation that we had about naming related to evm, we remember that each account created in `pallet-evm` already has code. So, what about 
`fn is_managed_by_evm(...)` and `fn all_accounts_managed_by_evm(...)`?

crates/humanode-runtime/src/lib.rs
```diff
@@ -874,13 +874,46 @@ pub type UncheckedExtrinsic =
...
+impl pallet_evm_system::migrations::broken_nonces_recovery::EvmStateProvider<EvmAccountId>
+    for EvmStateProvider
+{
+    fn has_code(account_id: &EvmAccountId) -> (bool, Weight) {
+        let has_code = pallet_evm::AccountCodes::<Runtime>::contains_key(account_id);
+        let weight = <Runtime as frame_system::Config>::DbWeight::get().reads(1);
+        (has_code, weight)
+    }
+
+    fn accounts_managed_by_evm() -> impl Iterator<Item = (EvmAccountId, Weight)> {
```
dmitrylavrenov, line 889: Should we keep explicit `Weight` value for each account instead of returning the full `Weight` for executed method ?

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// Execute migration to recover broken nonces.
+pub struct MigrationBrokenNoncesRecovery<EP, T>(sp_std::marker::PhantomData<(EP, T)>);
+
+#[cfg(feature = "try-runtime")]
+#[derive(Encode, Decode)]
+/// Key indicators of the state before runtime upgrade.
+struct PreUpgradeState {
+    /// Accounts' count.
+    accounts: u64,
```
dmitrylavrenov, line 31: Should we add check that account `data` value isn't changed during migration for each fixed account  managed by evm?

dmitrylavrenov: I can assume that it's redundant, but it's better to ask :)

crates/humanode-runtime/src/lib.rs
MOZGIII: I was thinking about this too - this API structure seems to loose the cost of weights that were filtered out, as it doesn't offer a meaningful way to carry them over.

MOZGIII:
Does the implementation have to live in a pallet? Cause if not - we can easily create a whole new create for this, and get rid of the indirection between the `pallet-evm-system` and `pallet-evm`.

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
quasiyoke: I'm for `evm-utils`. I also think it'll be useful after the second use case.

quasiyoke:
> Does the implementation have to live in a pallet? Cause if not - we can easily create a whole new create for this, and get rid of the indirection between the `pallet-evm-system` and `pallet-evm`.

I think better to put the migration in the pallet, since the pallet's state is its responsibility.
**dmitrylavrenov** (paraphrased): it would be more clear to other potential users of the pallet that they might need a migration if the migration was situated in it.

dmitrylavrenov:
> Does the implementation have to live in a pallet? Cause if not - we can easily create a whole new create for this, and get rid of the indirection between the `pallet-evm-system` and `pallet-evm`.

I prefer to keep it at `pallet-evm-system` as it allows having a good understanding and focus on history and reasons all changes here.

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
```diff
@@ -0,0 +1,188 @@
...
+                let (is_known_to_evm, w) = EP::has(&contract_id);
+                weight.saturating_accrue(w);
+                if is_known_to_evm {
+                    return false;
+                }
+                let has_state = <Account<T>>::contains_key(contract_id);
+                weight.saturating_accrue(T::DbWeight::get().reads(1));
+                !has_state
+            })
+            .expect("There must be unused nonce");
```
dmitrylavrenov, line 153: What about the case:

#### `step n`:
- `is_known_to_evm = true`

#### `step n+1`:
- `is_known_to_evm = false`
- `has_state = true`

It's not clear why should we skip it here instead of fixing the nonce state as a part of `min_nonce` logic? Could you please explain it?

dmitrylavrenov (approved):

crates/humanode-runtime/src/lib.rs
MOZGIII: This: https://github.com/humanode-network/humanode/pull/1433#issuecomment-2650144234

Allocating is bad, imo. There are ways around it - but I don't even want to start looking into them before we address the above.

MOZGIII:
The discoverability of this migration for the potential users is not a high priority aspect for us for now. High chances, are that in practice this is a mute point entirely.

I'm thinking this is the same situation as with the frontier's sufficinents hotfix.

---

The placement of the migration code together with the pallet vs placement of the migration in a separate crate is the real question. The trade-off there is an follows:
- we can colocate the migration and the pallet, and it makes sense ; but we must add an indirection layer via a trait, and it has to be implemented by someone somewhere
- if we don't colocate, we can directly depend on both crates and have a fully confined implementation, without indirection - but with poorer discoverability

Now, my list of issues with the trade-offs above:
- poor discoverability if we don't colocate the migration together with the pallet (that's I commented on above already)
- a need for a trait implementation somewhere if we do the colocation with the pallet - that it currently placed at our runtime - and that I don't really like, should be a part of the migration itself, especially since the logic there is non-trivial

That last point has too many drawbacks, that outweigh the benefits not taking the first route: even if we consider external users, for them it not easy provide an implementation in their own runtimes. Which brings us to having a crate to tie both pallets together anyway, which, in turn, brings in the question of why do we need the indirection in the first place and ultimately to moving the migration into a separate crate entirely.

Now, this problem above is relevant for us too - I'd like to avoid placing this temporary and non-trivial code in our runtime to throw it away some day with little trace - so let's maybe have a proper crate for it? This is my key point, but it drags the whole chain of derived conclusions I outlines above. There is an alternative there though - we can move only the trait impl from runtime into a crate, and keep the indirection. What do you think?

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
quasiyoke: Thanks for noticing this, the state can keep a deposit made to the address before the contract was created and we mustn't skip the nonce for "pre-funded contracts". Yellowpaper describes this scenario well ([equation (99)](https://ethereum.github.io/yellowpaper/paper.pdf#section.7)). The balance of the future contract should include this deposit (the `v'` in yellowpaper).

crates/humanode-runtime/src/lib.rs
quasiyoke: Refactored to:

```rs
fn accounts_managed_by_evm() -> impl Iterator<Item = (Option<AccountId>, Weight)> 
```

The signature may raise questions, but it is still quite obvious how to use it.

quasiyoke:
1. Moved migration to `evm-nonces-recovery` crate.
1. Added about 1.5% of block weight to migrations (506.70% of block weight total).

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -88,12 +90,11 @@ impl<R> Default for FrontierPrecompiles<R> {
...
-where
-    R: pallet_evm::Config,
-{
-    pub fn used_addresses() -> sp_std::vec::Vec<H160> {
-        sp_std::vec![
+impl<R: pallet_evm::Config> TypedGet for FrontierPrecompiles<R> {
+    type Type = Vec<H160>;
+
+    fn get() -> Self::Type {
+        [
```
dmitrylavrenov, line 97: What about 
- using a wrapper over `FrontierPrecompiles` to pass it into `MigrationBrokenNoncesRecovery` instead of doing changes for `FrontierPrecompiles` itself to be satisfied getting precompiled addresses list ?
or
- just use hardcoded list of ignored contracts inside `evm-nonces-recover` crate without affecting  `FrontierPrecompiles` ?

crates/evm-nonces-recovery/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+            })
+            .expect("There must be unused nonce");
+        info!("Account {id} minimal valid nonce is {nonce}");
+        (nonce.into(), weight)
+    }
+
+    /// Checks if such account exists in EVM.
+    fn is_known_to_evm(account_id: &H160) -> (bool, Weight) {
+        let has_code = AccountCodes::<R>::contains_key(account_id);
+        let weight = <R as frame_system::Config>::DbWeight::get().reads(1);
```
dmitrylavrenov, line 171: Doesn't look it redundant to have `(bool, Weight)` for function signature when we always know that the function itself takes 1 read all time during execution? Can't we just properly calculate it in place where the function is called taking into account the function weight ?

Additionally, I would like to note that we can somewhere to do 1 READ for getting `account_codes` by `AccountCodes::<R>::get` and then use it without reading storage again

dmitrylavrenov:
> 1. Moved migration to `evm-nonces-recovery` crate.
> 2. Added about 1.5% of block weight to migrations (506.70% of block weight total).

The heavy part comes from offences removal migration

crates/pallet-evm-system/src/migrations/broken_nonces_recovery.rs
quasiyoke: ✅

crates/humanode-runtime/src/frontier_precompiles.rs
quasiyoke: Thought about using `FrontierPrecompiles::used_addresses` as const generic param, but const generics can only be primitive types. Added `FrontierPrecompilesAddresses` wrapper.

crates/evm-nonces-recovery/src/lib.rs
quasiyoke: Discussed it and decided that decomposition in this case would rather cause more confusion than good.

dmitrylavrenov (approved):
Looks good!

In my opinion, the current version is better. Just left https://github.com/humanode-network/humanode/pull/1433#pullrequestreview-2607987686 as unresolved to be discussed together

crates/evm-nonces-recovery/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        account.nonce = nonce;
+        weight
+    }
+
+    /// Computes the minimum possible nonce for a given account.
+    fn min_nonce(id: &H160) -> (<R as pallet_evm_system::Config>::Index, Weight) {
+        let nonce = (1..)
+            .find(|&nonce| {
+                let contract_id = contract_address(id, nonce);
+                let is_known_to_evm = AccountCodes::<R>::contains_key(contract_id);
```
MOZGIII, line 153: Note this won't give us 100% accurate results because of self-destructs. If a contract was created by then self-destructed, given the very bug we're fixing here, there will be no trace of it ever existing.

I suggest documenting it in the code as well.

---

That said, maybe we should apply some heuristics here, and only commit to a given nonce once we see a consecutive gap in absent contract addresses?

crates/evm-nonces-recovery/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+            return Default::default();
+        }
+        info!("Account {id} has zero nonce");
+        let (nonce, weight) = Self::min_nonce(id);
+        account.nonce = nonce;
+        weight
+    }
+
+    /// Computes the minimum possible nonce for a given account.
+    fn min_nonce(id: &H160) -> (<R as pallet_evm_system::Config>::Index, Weight) {
```
MOZGIII, line 149: We want to add a (*possibly* configurable, but may be hard-coded) hard-cap on the iterations this fn allows to do; the rationale is we'd want to avoid a situation where, if some of our assumptions fails (i.e. about how the `AccountCodes::<R>::contains_key` works) we won't exhaust the block weight.

In such cases, it might be better to abort the migration altogether and demand immediate attention from the devs.

Triggering a panic would not make sense for this, however, since it would be effectively the same as exhausting the block weight - the chain won't be able to progress. This means a log with an error result + a short-cricut early return at the callsite is the way to implement this.

crates/evm-nonces-recovery/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// - If the state is missing, recreates it.
+    /// - If the state has nonce = 0, writes the smallest possible valid nonce.
+    ///
+    /// Precompiled contracts in Ethereum usually have nonce = 0 as well. However, since precompiled
+    /// contracts are typically implemented by hooking calls to specific addresses and adding dummy
+    /// state (to ensure they are callable like regular contracts), there's no need for a non-zero
+    /// nonce unless they explicitly perform state-changing operations like `CREATE`. Therefore,
+    /// precompiled contracts MUST NOT be passed here.
+    fn recover(id: &H160, account: &mut Option<AccountInfoOf<R>>) -> Weight {
+        let Some(account) = account.as_mut() else {
```
MOZGIII, line 130: Note this branching could be unified if we do a `let account = account.get_or_insert_with(Default::default());` (except for the logging there).

MOZGIII (approved):
Looks great!

There are a couple of comments I've left, but all of them are just nits. This is good stuff.

crates/humanode-runtime/src/lib.rs
MOZGIII: Just for the sake of bringing this point up: in some cases, we'd want to iterate over a key prefix across the multiple blocks; i.e. maintain a cursor to track how much work we already did vs how much work is still left to do. This is often the case when the key space of the prefix is large enough that it would cause trouble if we iterate over it in just one sweep.

For such cases, an API would require a way to communicate the weight of accessing the cursor.

So, ultimately, we might want to have an iterator that can also carry an accumulated weight alongside of it that is no directly associated with the weight of each particular iteration. For such cases, we could use the similar signature, and just return the `(impl Iterator<Item = (Option<AccountId>, Weight)>, Weight)`- to encode the initialization weight - but that'd be a bit too verbose...

Anyhow - I like the API we have now. It models the underlying operations properly.

As a different and more universal approach, we could have a `fn accounts_managed_by_evm(weight_accumulator: &mut Weight) -> impl Iterator<Item = AccountId>` as a more natural way of modeling the weight accumulation with the borrow checker rules - but the current approach makes it even more explicit when we are doing the steps, and that each step has a deterministic weight associated with it. So, for the present case, and for how we ended up setting up the code - the current API is more expressive. However, for something to use in the traits and with indirection, I think passing in the `&mut Weight` for the accumulated weight collection is the way to go.

crates/evm-nonces-recovery/src/lib.rs
quasiyoke: ✅ Added `MaxNonceGuesses` parameter

crates/evm-nonces-recovery/src/lib.rs
quasiyoke: > If a contract was created by then self-destructed, given the very bug we're fixing here, there will be no trace of it ever existing.

Added a comment to `fn restore`:

```rs
/// Bug #1402 made it possible for "Self-destruct" to delete not only the contract code,
/// but also its state with the nonces used. We have no way to restore their nonces here,
/// but in the future, contracts won't be created at such addresses anyway.
```

> That said, maybe we should apply some heuristics here, and only commit to a given nonce once we see a consecutive gap in absent contract addresses?

I haven't made any efforts in this direction yet.

MOZGIII: You think it's necessary? If not we can wrap it up.

dmitrylavrenov (approved):
LGTM!

crates/evm-nonces-recovery/src/lib.rs
MOZGIII: While it is possible in theory, we can only care about the list of cases we have at hand, and inspect all them for any self-destructs.

crates/evm-nonces-recovery/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+                error!("Unable to compute nonce for account {id}");
+                return weight;
+            }
+        };
+        account.get_or_insert_with(Default::default).nonce = nonce;
+        weight
+    }
+
+    /// Computes the minimum possible nonce for a given account.
+    fn min_nonce(id: &H160) -> Result<(<R as pallet_evm_system::Config>::Index, Weight), Weight> {
```
MOZGIII, line 157:
```suggestion
    ///
    /// # Note
    ///
    /// This routine might not be accurate.
    ///
    /// We had a bug which would cause a self-destructed contract to clear out all of it's state, instead of just clearing up the code and leaving the nonce intact. This migration is the very thing that aims to fix this situation.
    /// This explains why this fn is implemented by just looking at the [`AccountCodes`], and not considering
    /// account balance or nonce to check the account existence - but either way would work here.
    ///
    /// However! Due to above, this routing might not be actually finding the nonces suitable for recovery.
    /// Here we stop at the first "unused nonce", but given that self-destruct-induced
    /// gaps in the allocated contract addresses are possible, this code may end up giving an nonce that has no contract address, yet is not at the end of the span of all the allocated contract addesses but in the middle of it (in the gap). Recovering to this nonce would not really work, and only allow for one (or however long the gap sequence is) contract to be created.
    ///
    /// A better implementation would be to scan for the gaps and fill them with dummy contracts entries occupying the gap (matching the outcome of a state that a self-destructed account would have, if we didn't have a bugged code) in addition to fixing the creating contract nonce.
    fn min_nonce(id: &H160) -> Result<(<R as pallet_evm_system::Config>::Index, Weight), Weight> {
```

Here's the documentation I had in mind for https://github.com/humanode-network/humanode/pull/1433/files#r1959369497

quasiyoke: ✅ pasted your comment, thanks! Summarizing a little below:

To make sure that `min_nonce` hasn't been used before (and the account at that address hasn't been self-destructed), it would make sense to additionally check:
- Guess N nonce values following `min_nonce`. For example:
  - N = `total_accounts_count`, or
  - N = Σ nonces (for all accounts)
- Try to recursively compute `min_nonce` for addresses that appear unoccupied (it may turn out that there's code on addresses formed from them). The depth of recursion would also be worth discussing.

This amount of computation is difficult to perform in our migration execution model so we opened a ticket https://github.com/humanode-network/humanode/issues/1448

PR #1434, opened 2025-02-10
Clear all storage items associated with removed Offences pallet at runtime level

dmitrylavrenov:
Done using [RemovePallet](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/support/src/migrations.rs#L298)

`try-runtime` run over mainnet:
```
2025-02-10 17:57:06.354  INFO main frame_support::migrations: Found Offences keys pre-removal 👀    
2025-02-10 17:57:06.481  INFO main frame_support::migrations: Removed 80814 Offences keys 🧹    
2025-02-10 17:57:06.819  INFO main frame_support::migrations: No Offences keys found post-removal 🎉 
```

PR #1444, opened 2025-02-18
Revert "Clear all storage items associated with removed Offences pallet at runtime level (#1434)"

dmitrylavrenov:
My bad, missed that consumed weight is much higher than total weight during `try-runtime` run over mainnet (over `500%`).

After revert its about `0.02%`


quasiyoke (approved):

dmitrylavrenov:
Planning to be implemented with https://github.com/humanode-network/humanode/issues/1447

PR #1446, opened 2025-02-18
Make Native-to-EVM Bridge transactions visible via Ethereum RPC

dmitrylavrenov:


dmitrylavrenov:
Discussed - it's better to introduce new currency swap logic explicitly related ethereum

Issue #1004, opened 2024-04-17
Consider reimplementing currncy-swap using standardtized asset-rate substrate functionality


MOZGIII:
https://github.com/paritytech/substrate/commit/bbcc63a96c8b77b9d07968cc2612a5c15b8cc0e5
https://github.com/paritytech/substrate/commit/917f7e38c58cbf98ef2a2ed460bab530b160bc1e

PR #1449, opened 2025-02-19
Implement ethereum explicitly related currency swap logic to make Native-to-EVM Bridge transactions visible via Ethereum RPC

dmitrylavrenov:
## Design thoughts

- A new pallet that contains ethereum explicitly related currency swap logic. We need to make Native-to-EVM Bridge transactions visible via Ethereum RPC
- Use https://github.com/humanode-network/frontier/blob/174c0227af00dbc26485c1eac68738e1d01980e8/frame/ethereum/src/lib.rs#L964 to make the transaction visible at ethereum side
- Avoid using `Currency` in favour of fungible related traits like `Inspect` and `Mutate`

## Tasks

- [x] pallet implementation
- [x] precompile implementation
- [x] runtime integration
- [x] runtime unit tests
- [x] unit tests for native to evm swap
- [x] unit tests for evm to native swap
- [x] e2e tests for swap in both directions

P.S: do benchmarks at separate PR

### transaction itself

Screenshot 2025-02-19 at 23 01 35

### transaction execution in explorer on Substrate RPC side

Screenshot 2025-02-19 at 23 02 22

### transaction execution in explorer on Ethereum RPC side

[Very tiny](https://github.com/Alethio/ethereum-lite-explorer) ethereum explorer is used to verify visibility via Ethereum RPC

Screenshot 2025-02-19 at 23 02 43


crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+                .into_result()?;
+
+            T::NativeCurrency::transfer(&who, &T::PotNativeBridge::get(), amount, preservation)?;
+
+            let evm_balance_to_be_deposited: u128 =
+                estimated_swapped_balance.unique_saturated_into();
+
+            let transaction = pallet_ethereum::Transaction::EIP1559(ethereum::EIP1559Transaction {
+                chain_id: <T as pallet_evm::Config>::ChainId::get(),
+                nonce: pallet_evm::Pallet::<T>::account_basic(&T::PotEvmBridge::get().into())
```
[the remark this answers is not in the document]
dmitrylavrenov: What do you mean, could you explain, please ?

In current version we use `PotEvmBridge` nonce as a source transfer of tokens as it's a real transfer from `PotEvmBridge` to `Target` address. In case transaction is executed => nonce is increased by 1.

Checked the flow by executing currency swap. `trx_hash` is different each time, nonce is increased.

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+                    .0
+                    .nonce,
+                max_priority_fee_per_gas: 0.into(),
+                max_fee_per_gas: <T as pallet_evm::Config>::FeeCalculator::min_gas_price().0,
```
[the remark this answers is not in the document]
dmitrylavrenov: Yep, we can. Set to zero.

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+                s: Default::default(),
+            });
+
+            let evm_trx_hash = transaction.hash();
+
+            let _post_info = pallet_ethereum::ValidatedTransaction::<T>::apply(
+                T::PotEvmBridge::get().into(),
+                transaction,
+            )
+            .map_err(|dispatch_error_with_post_info| dispatch_error_with_post_info.error)?;
```
dmitrylavrenov, line 172: Should we take into account `post_info` as `apply` return `DispatchResultWithPostInfo`, adjust it with native transfer info and return `DispatchResultWithPostInfo` instead of `DispatchResult` at the calls ?

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+            let estimated_swapped_balance = T::BalanceConverter::convert(amount);
+            T::EvmCurrency::can_deposit(&to, estimated_swapped_balance, Provenance::Extant)
+                .into_result()?;
+
+            T::NativeCurrency::transfer(&who, &T::PotNativeBridge::get(), amount, preservation)?;
+
+            let evm_balance_to_be_deposited: u128 =
+                estimated_swapped_balance.unique_saturated_into();
+
+            let transaction = pallet_ethereum::Transaction::EIP1559(ethereum::EIP1559Transaction {
```
dmitrylavrenov, line 149: Should we move transaction construction to runtime level and use some interface here instead ?

crates/pallet-native-to-evm-currency-swap/src/lib.rs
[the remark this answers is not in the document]
MOZGIII: Ok, this makes sense, great!

crates/pallet-native-to-evm-currency-swap/src/lib.rs
MOZGIII: No, why?

dmitrylavrenov: Just a little question that I had in mind, I prefer to leave it here, but it's better to reask )

MOZGIII: Btw, can we use the `..Default::default()` notation here?

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+            /// The account id balances withdrawed from.
+            from: T::AccountId,
+            /// The withdrawed balances amount.
+            withdrawed_amount: NativeBalanceOf<T>,
+            /// The account id balances deposited to.
+            to: T::EvmAccountId,
+            /// The deposited balances amount.
+            deposited_amount: EvmBalanceOf<T>,
+            /// The corresponding transaction hash executed in evm.
+            evm_trx_hash: H256,
```
MOZGIII, line 93:
```suggestion
            evm_transaction_hash: H256,
```

crates/pallet-native-to-evm-currency-swap/src/lib.rs
MOZGIII: I don't know. It looks like the bench would take care of this, no?

MOZGIII: In short: we'd need to bench something, and it will depend on our bench structure. As the code is right now, the bench will cover all of it, and there's no need to adjust the post info.

MOZGIII (changes requested):
This looks good!

If we add the E2E tests to verify that the solution works overall, and unit tests with mock runtime (as we usually do) to test the niche cases (zero transfer, etc) and for tight regression control - I think we ship it pretty much right away

crates/pallet-native-to-evm-currency-swap/src/lib.rs
MOZGIII: Rationale: this `trx` abbreviation is competing with `tx`, and I'd rather avoid the hassle of picking one at all.

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+                estimated_swapped_balance.unique_saturated_into();
+
+            let transaction = pallet_ethereum::Transaction::EIP1559(ethereum::EIP1559Transaction {
+                chain_id: <T as pallet_evm::Config>::ChainId::get(),
+                nonce: pallet_evm::Pallet::<T>::account_basic(&T::PotEvmBridge::get().into())
+                    .0
+                    .nonce,
+                max_priority_fee_per_gas: 0.into(),
+                max_fee_per_gas: 0.into(),
+                gas_limit: 21000.into(), // simple transfer
```
MOZGIII, line 156: This should be taken from the `EvmConfig` or whatever that is called; the rationale is this value can change across EVM versions (i.e. London vs Istambul, that sort of thing).
Either expose this as a config parameter of the pallet itself, or just directly instantiate from the EVM config.

crates/pallet-native-to-evm-currency-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// The converter to determine how the balance amount should be converted from native
+        /// to evm currency.
+        type BalanceConverter: Convert<NativeBalanceOf<Self>, EvmBalanceOf<Self>>;
+
+        /// The pot native bridge account.
+        type PotNativeBridge: Get<Self::AccountId>;
+
+        /// The pot evm bridge account.
+        type PotEvmBridge: Get<Self::EvmAccountId>;
```
MOZGIII, line 73:
```suggestion
        /// The pot native bridge account.
        type BridgePotNative: Get<Self::AccountId>;

        /// The pot evm bridge account.
        type BridgePotEvm: Get<Self::EvmAccountId>;
```

This makes more sense actually, wouldn't you say?

crates/pallet-native-to-evm-currency-swap/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "pallet-native-to-evm-currency-swap"
```
MOZGIII, line 2:
```suggestion
name = "pallet-native-to-evm-swap"
```

We can actually drop the "currency" from the name of this pallet, as it no longer requires it to be implemented

crates/pallet-native-to-evm-currency-swap/src/lib.rs
dmitrylavrenov: Done - https://github.com/humanode-network/humanode/pull/1449/commits/c68cf72734fa580ce72a01d61d4a3a96b54c8639

crates/pallet-native-to-evm-currency-swap/src/lib.rs
dmitrylavrenov: Yep, done - https://github.com/humanode-network/humanode/pull/1449/commits/d45cf119c11a0a84b5417dc3bba6da303cdf2c8b

crates/pallet-native-to-evm-currency-swap/Cargo.toml
dmitrylavrenov: Got it, done - https://github.com/humanode-network/humanode/pull/1449/commits/4c9425cb193ebcace3e7bddfb39b205262ddefc2

crates/pallet-native-to-evm-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// The EVM user account identifier type.
+        type EvmAccountId: Parameter
+            + Member
+            + MaybeSerializeDeserialize
+            + Debug
+            + MaybeDisplay
+            + Ord
+            + MaxEncodedLen
+            + Into<H160>;
```
MOZGIII, line 57: Do we need all those bounds?

dmitrylavrenov: Will recheck

crates/pallet-native-to-evm-swap/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// Native currency.
+        type NativeCurrency: Inspect<Self::AccountId> + Mutate<Self::AccountId>;
+
+        /// EVM currency.
+        type EvmCurrency: Inspect<Self::EvmAccountId>;
```
MOZGIII, line 63:
```suggestion
        /// Native currency.
        type NativeToken: Inspect<Self::AccountId> + Mutate<Self::AccountId>;

        /// EVM currency.
        type EvmToken: Inspect<Self::EvmAccountId>;
```

In the interest of moving away from currency - I think this naming change makes sense. Or should we use `Fungible`, or keep as-is instead? Let's think about it?

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: Thanks, removed redundant - https://github.com/humanode-network/humanode/pull/1449/commits/04390be3d3721028a367817ea06bb4be42ee1a3c

crates/pallet-native-to-evm-currency-swap/src/lib.rs
dmitrylavrenov: Extracted `gas_limit` value from `EvmConfig` - https://github.com/humanode-network/humanode/pull/1449/commits/f48f89ea108e58aff831a9f014459b0795e0ac41

crates/pallet-native-to-evm-swap/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                access_list: Default::default(),
+                odd_y_parity: false,
+                r: Default::default(),
+                s: Default::default(),
+            });
+
+            let evm_transaction_hash = transaction.hash();
+
+            let _post_info = pallet_ethereum::ValidatedTransaction::<T>::apply(
+                T::BridgePotEvm::get().into(),
```
MOZGIII, line 163: Do we want to call this tx on behalf of the pot itself, or maybe somehow organize it such that the precompile issues the transfer?

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: Good suggestion. In my opinion, token option sounds better as we usually talk about `HMND` token and `eHMND` token. One more point that we are moving away from currency.

MOZGIII:
Thinking about this, we should explicitly test, among other things, that the *pot* **doesn't** *pay the fee* for any tx with this flow.

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1449/commits/9ffac3469a27e3614780933659c4c1fa78179e5b

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: `T::BridgePotEvm` is used here as a source of the transaction. It means, that tokens are transferred from source to target. I think it's a right usage to make it visible at it happens.

MOZGIII: It is true, but it doesn't have to be the bridge pot that issues the tx, or does it? Our precompile that moves the funds in the opposite direction, for instance, does not reveal that it transfers the funds to the pot.

In my view, it would make sense to actually transfer the funds as if they come from the precompile, at least at the very high level of thinking (i.e. non-technical).
So, the question is does it make sense; and if it does - what would it cost us to implement it; if it doesn't make sense - how do we see the lack of symmetry between the fact that the funds are received from one address (the pot) when they are swapped to EVM, but when you swap them from EVM you have to do it via a different address (via a precompile).

dmitrylavrenov: Just one question, it's not just a behalf. We pass here some account that is used like source of tokens in the following way:
-  https://github.com/humanode-network/frontier/blob/931a70882b3c095c1b7bcd6adea0f66ee24ca86b/frame/ethereum/src/lib.rs#L991
- https://github.com/humanode-network/frontier/blob/931a70882b3c095c1b7bcd6adea0f66ee24ca86b/frame/ethereum/src/lib.rs#L600
- https://github.com/humanode-network/frontier/blob/931a70882b3c095c1b7bcd6adea0f66ee24ca86b/frame/ethereum/src/lib.rs#L823

So, if we pass another account then it should have the same amount of tokens target account expect.

Is your idea that we should somehow send tokens to precompile first and then use precompile as a source ?

MOZGIII: Technically, we can't just do it - the precompile has the code, so it can't possible issue a transaction. So, I guess that settles it.

MOZGIII: Actually... If we make a call *into* a precompile *from* the pot address, passing the swap `value` along and - obviously this won't be a simple transfer anymore - the address of the actual swap recipient in the calldata; in other words we'd be calling a particular smart contract function on the precompile.
Then what we could do is check the validity of the operation at the precompile side - i.e. reject calls from any accounts but the pot account - and then we just transfer the received `value` to the swap recipient.

crates/pallet-native-to-evm-swap/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+                input: Default::default(),
+                access_list: Default::default(),
+                odd_y_parity: false,
+                r: Default::default(),
+                s: Default::default(),
+            });
+
+            let evm_transaction_hash = transaction.hash();
+
+            let _post_info = pallet_ethereum::ValidatedTransaction::<T>::apply(
```
MOZGIII, line 161: To revisit the `PostDispatchInfo` question; turns out (after inspection of the `pallet_ethereum::ValidatedTransaction` implementation) the `post_info` is highly non-trivial; it has `pays: No`, weight based on the EVM gas consumed (which is, nonetheless, covered by benches if we bench this call).

So, maybe we should, at least, apply some scrutiny to the returned value, instead of just `_post_info`. A `debug_assert!` would be a good idea here.

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: Just to think as a second approach.

What about considering implementing in the same way `precompile-evm-to-native-swap` that actually will do:
- transfer from `evm source` to `BridgePotEvm`
- transfer from `BridgePotNative` to `native target`

Then we will have the same tokens flow in both direction of the swap.

dmitrylavrenov: Btw, I agree that having the same tokens flow in both direction would be good design.

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: That's why I asked before, good suggestion that we can add `debug_assert` to verify that we consume expected gas, added - https://github.com/humanode-network/humanode/pull/1449/commits/067f1c39f22811decc858ab472a064c1f474494a

dmitrylavrenov: Moved ethereum transfer logic to separate fn - https://github.com/humanode-network/humanode/pull/1449/commits/eaf3920eb2ab819020d6cf7ab377222d4b1c4c72

crates/pallet-native-to-evm-swap/src/lib.rs
dmitrylavrenov: Just was thinking about it - https://github.com/humanode-network/humanode/pull/1449#discussion_r1963890013. Is it allowed to be the swap tokens flow visible for the target address. In which way we can say it's visible as target address isn't used as a destination ?

MOZGIII: > Btw, I agree that having the same tokens flow in both direction would be good design.

Emphasis on the same tokens here? Or do you mean flowing through the "same route" as in from and to the precompile address?

MOZGIII: > Just to think as a second approach.
> 
> What about considering implementing in the same way `precompile-evm-to-native-swap` that actually will call precompile to do the following steps:
> 
> * transfer from `evm source` to `BridgePotEvm`
> * transfer from `BridgePotNative` to `native target`
> 
> Then we will have the same tokens flow in both direction of the swap.

Isn't this exactly what we're doing now and also did before? I don't get it...

dmitrylavrenov: Currently, we use the following flow at `precompile-currency-swap`:
- `source evm address` calls `precompile` to do swap of `N eHMND` tokens to `target native address`
- `precompile` just withdraws `N eHMND` tokens from `source evm address` (**not etherem transfer**) => precompile has `eHMND` imbalance
- `precompile` does `CurrencySwapT` getting `HMND` imbalance
- `precompile` does `resolve_creating` to `native target address` (**not native transfer**)

MOZGIII: > Just was thinking about it - [#1449 (comment)](https://github.com/humanode-network/humanode/pull/1449#discussion_r1963890013). Is it allowed to be the swap tokens flow visible for the target address. In which way we can say it's visible as target address isn't used as a destination ?

What do you mean? Ideally we'd want our swap system to look from the outside as if the user is interacting with some clearly system-level contract (precompile with a neatly looking address in our case).
The main way users will observe those interactions is via the explorer, as other visible things - like the code and RPC-exposed APIs will be handled by the devs.

MOZGIII: > Currently, we use the following flow at `precompile-currency-swap`:
> 
> * `source evm address` calls `precompile` to do swap of `N eHMND` tokens to `target native address`
> * `precompile` just withdraws `N eHMND` tokens from `source evm address` (**not etherem transfer**) => precompile has `eHMND` imbalance
> * `precompile` does `CurrencySwapT` getting `HMND` imbalance
> * `precompile` does `resolve_creating` to `native target address` (**not native transfer**)

That last part is fine, because we don't care about it as much since it all has to be accounted by the explorer somehow, and they do it pretty well.

However, the **not etherem transfer** at the second step concerns me - we should probably make it a proper EVM transfer at the precompile side too. In that sense, we could implement a whole new precompile too as part of this work we're doing now.

dmitrylavrenov: > However, the not etherem transfer at the second step concerns me - we should probably make it a proper EVM  transfer at the precompile side too. In that sense, we could implement a whole new precompile too as part of this work we're doing now.

Absolutely agree on it!

dmitrylavrenov: @MOZGIII So, I'm going to implement in the same way precompile logic.

Before it, I would like to finalize on exact ethereum part flow. Should either we have a special precompile call that you described here - https://github.com/humanode-network/humanode/pull/1449#discussion_r1963890013 or we use just simple transfer with bridge pot accounts as a source/target in ethereum transaction ?

dmitrylavrenov: Discussed the approach with precompile and agreed on using just simple transfer with bridge pot accounts as a source/target in ethereum transaction. The main reasons we would like to allow users directly see swap evm target address as a part of ethereum transaction instead of having it at calldata.

dmitrylavrenov: @MOZGIII please, check current design implementation. If it's ok I will continue working on thorough testing environment, unit tests, e2e tests

crates/humanode-runtime/src/frontier_precompiles.rs
```diff
@@ -76,6 +77,8 @@ pub mod precompiles_constants {
     pub const NATIVE_CURRENCY: u64 = 2050;
     /// `CurrencySwap` precompile constant.
     pub const CURRENCY_SWAP: u64 = 2304;
```
MOZGIII, line 79: We'll have to eventually deprecate this one; how would we do it?

MOZGIII (approved):

crates/pallet-native-to-evm-swap/src/lib.rs
MOZGIII: This is great!

crates/humanode-runtime/src/frontier_precompiles.rs
dmitrylavrenov: Yep, the interfaces are compatible. So, just remove the old one and leave the same precompile code ?

dmitrylavrenov:
Created a follow up issue to apply evm swap design and fungible interfaces at primitives currency swap related implementations - https://github.com/humanode-network/humanode/issues/1472

crates/pallet-evm-swap/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = "pallet-evm-swap"
```
MOZGIII, line 2:
```suggestion
name = "pallet-native-to-evm-swap"
```

I think it makes more sense if we'll be splitting the crates.

dmitrylavrenov: Got it. Should we call precompile logic as `precompile-evm-to-native-swap` ? Or, just `precompile-evm-swap` is enough ?

MOZGIII: `precompile-evm-to-native-swap` makes sense for symmetry

crates/humanode-runtime/src/frontier_precompiles.rs
dmitrylavrenov: Done

crates/pallet-evm-swap/Cargo.toml
dmitrylavrenov: Done

quasiyoke (commented):
Reviewed, great job!

As far as I can tell, `pallet-currency-swap` and `pallet-native-to-evm-swap` have identical Substrate interface. The general layout of pallets was covered in the discussions [[1]](https://github.com/humanode-network/humanode/pull/1449#pullrequestreview-2638277325) and [[2]](https://github.com/humanode-network/humanode/pull/1449#discussion_r1980390855) — but I don't understand:

1. Why we are making a new `pallet-native-to-evm-swap` instead of modifying 1pallet-currency-swap`?
1. Will we edit our EVM-playground to use `pallet-native-to-evm-swap`?
   1. Should we remove `pallet-currency-swap` after edits in EVM-playground?
1. Should we close ticket #1004

crates/humanode-runtime/src/evm_swap.rs
```diff
@@ -1,4 +1,5 @@
 use bridge_pot_currency_swap::ExistenceRequired;
+use precompile_evm_to_native_swap::Config as PrecompileConfig;
```
MOZGIII, line 2: Don't do this

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1449/commits/5c39139c3cb2d10735ed44642f6b454e96a35b03

dmitrylavrenov:
> 1. Why we are making a new `pallet-native-to-evm-swap` instead of modifying 1pallet-currency-swap`?

Yep, the pallets have identical Substrate interface in terms of calls but not compatible in terms of API to submit such extrinsics. We can't just rename existing pallet as it might break others running tools that work with existing swap due to using `currencySwap.swap(...)` call on their end. We need to wait when they switch to `nativeToEvmSwap.swap(...)`.

We don't want to leave general currency note in `pallet-currency-swap` naming that means any tokens based on currency behaviour can be used here. Instead, we would like to explicitly note that it's a logic to swap native tokens with EVM based tokens containing some ethereum related implementations. 

> 2. Will we edit our EVM-playground to use `pallet-native-to-evm-swap`?
>    
>    1. Should we remove `pallet-currency-swap` after edits in EVM-playground?

Sure, as I mentioned before, all tools that works with swap should switch to a new native to EVM swap logic. Obviously, we are going to deprecate it in our runtime. Additionally, I guess, it will be removed from our codebase later as currency behaviour is going to be deprecated at polkadot-sdk.

> 3. Should we close ticket [Consider reimplementing currency-swap using standardtized asset-rate substrate functionality #1004](https://github.com/humanode-network/humanode/issues/1004)

I don't think that this PR should close this ticket as it doesn't reimplement `pallet-currency-swap`. But, I guess, the ticket can be closed as _not relevant_ when we deprecate `pallet-currency-swap`.

quasiyoke (approved):

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
```diff
@@ -0,0 +1,191 @@
...
+      if (
+        item.event.section == "transactionPayment" &&
+        item.event.method == "TransactionFeePaid"
+      ) {
+        transactionPaymentEvent = item.event as unknown as IEvent<
+          Codec[],
+          TransactionPaymentEvent
+        >;
+      }
+    }
```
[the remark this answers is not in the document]
dmitrylavrenov: I believe as well ) but it's smth that I were able to make it works. 

Could you suggest a better way here ?

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
```diff
@@ -0,0 +1,191 @@
...
+  beforeEachWithCleanup(async (cleanup) => {
+    node = runNode({ args: ["--dev", "--tmp"] }, cleanup.push);
+
+    await node.waitForBoot;
+
+    substrateApi = await substrate.apiFromNodeWebSocket(node, cleanup.push);
+    ethPublicClient = eth.publicClientFromNodeWebSocket(node, cleanup.push);
+  }, 60 * 1000);
+
+  it("success", async () => {
```
MOZGIII, line 42: I don't see where we test the visibility on the EVM side; should be a check where we use `ethPublicClient` to get the block and check that there is a tx with the expected txhash in it and expected `to`, `from`, `value` and `data` (`0x`). This is what we're after.

dmitrylavrenov: There is a check on ethereum execution event containing the expected evm transaction hash.

Btw, thanks for noting, will add more details including mentioned fields

MOZGIII (changes requested):
The crucial test is missing

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
MOZGIII: > There is a check on ethereum execution event containing the expected evm transaction hash.

Yes, but that's not really where all those systems that work with EVM RPCs look at. So that's good, but not really relevant for what we're trying to solve here - and there are a lot of ways this can go wrong even if the event at substrate side exists.

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1449/commits/45086a6358479c99af1a94ecaaa7d154ffbb287a

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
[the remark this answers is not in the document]
MOZGIII: Two ways:
1. Ensure we actually get the events we are looking for; currently, if the event is not present, this code won't error out.
2. Ensure correct events order - this is an even stricter check, but different in nature from expecting a particular set of events.

Both 1 and 2 combined can be implemented by checking that the whole events array matches the expectation.

MOZGIII (commented):
The essential check is not present, great!

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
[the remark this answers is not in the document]
dmitrylavrenov: About 1 we have asserts that we got all events we are looking for:

```
    assert(nativeToEvmSwapBalancesSwappedEvent);
    assert(ethereumExecutedEvent);
    assert(transactionPaymentEvent);
```

https://github.com/humanode-network/humanode/blob/9966d50842dcf5711cbf65f8a63cd716d71cf468/utils/e2e-tests/ts/tests/swap/nativeToEvm.ts#L114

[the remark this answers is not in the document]
dmitrylavrenov: Btw, I guess, checking all executed events and their order sounds better. Will try to implement it.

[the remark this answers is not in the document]
MOZGIII: Oh, right; I didn't read the code fully, just skimmed though; the way it is written you got to look at, sort of, two places - which is not great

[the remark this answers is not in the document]
dmitrylavrenov: I've come up with the following way in ts - https://github.com/humanode-network/humanode/pull/1449/commits/79452fbb41a4d2103845c7ce5a268a9602886ae1. Does it look good?

crates/precompile-evm-to-native-swap/EvmToNativeSwap.sol
```diff
@@ -19,4 +19,14 @@ interface CurrencySwap {
...
+
+  /**
+   * Event emitted when a transfer has been performed.
+   * Selector: 69d31d1d87c1206beee49fbab11570a7f001121cf21fbb234d5b0a2473fa5c58
+   *
+   * @param from The EVM account id the tokens withdrawed from.
+   * @param to The Native account id the tokens deposited to.
+   * @param value The amount of tokens swapped.
+   */
+  event Swap(address indexed from, bytes32 indexed to, uint256 value);
```
MOZGIII, line 31: To do; check if the name `Swap` in meaningful here

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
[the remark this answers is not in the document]
dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1449/commits/826c5b1703c61507a8efc206677e021627ab5926

crates/precompile-evm-to-native-swap/EvmToNativeSwap.sol
dmitrylavrenov: Didn't get. It's checked here - https://github.com/humanode-network/humanode/blob/826c5b1703c61507a8efc206677e021627ab5926/utils/e2e-tests/ts/tests/swap/evmToNative.ts#L88

dmitrylavrenov: I've discovered that we missed it when decided to walk through `precompile-native-currency` to recheck precompile code structure.

MOZGIII: I mean, I left this note for myself to remind to check if having it called `Swap` vs say `Swapped` makes sense

crates/pallet-native-to-evm-swap/src/tests.rs
MOZGIII, line 1: Should we test for events here?

crates/precompile-evm-to-native-swap/src/tests.rs
```diff
@@ -0,0 +1,497 @@
...
+            Vec::new(),
+            true,
+            true,
+            None,
+            None,
+            <Test as pallet_evm::Config>::config(),
+        )
+        .unwrap_err();
+        assert!(matches!(execerr.error, pallet_evm::Error::BalanceLow));
+        assert_eq!(
```
MOZGIII, line 419: Maybe use an `assert_noop` or something?

crates/pallet-native-to-evm-swap/src/tests.rs
dmitrylavrenov: It's tested at https://github.com/humanode-network/humanode/blob/fa2238fa899cb468984b17a305c93ea540bdcd21/crates/pallet-native-to-evm-swap/src/tests.rs#L100 during `run_succeded_test_and_assert`

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
```diff
@@ -0,0 +1,213 @@
...
+      targetSwapEvmAddress,
+    );
+    expect(ethereumExecutedEvent.data.exitReason.toPrimitive()).toEqual({
+      succeed: "Stopped",
+    });
+
+    // Balances changes related checks.
+    const fee = BigInt(
+      transactionPaymentEvent.data.actualFee.toPrimitive() as unknown as bigint,
+    );
```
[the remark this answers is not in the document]
dmitrylavrenov: Yep, but it's smth that works. Would be happy to learn more elegant solutions. I were not able to come up with smth better.

MOZGIII (approved):
This is good, left a few notes but they're minor

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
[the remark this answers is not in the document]
MOZGIII: The return type of `.toPrimitive()` is `AnyJson`, which is this:
```ts
export type AnyJson = string | number | boolean | null | undefined | AnyJson[] | {
    [index: string]: AnyJson;
};
```

Here, we know we expect this to be a numeric value, so it would either be a `number` or a `string` (for when the number doesn't fit info a standard JS "float" number).

On the other hand, `BigInt` constructor function is this: `(value: bigint | boolean | number | string): bigint;`.

So, if we tell TypeScript to narrow the types accordingly via `.toPrimitive() as string | number` instead of rebinding them entirely via `as unknown as ...` we still have some level of type safety.

```suggestion
    const fee = BigInt(
      transactionPaymentEvent.data.actualFee.toPrimitive() as string | number,
    );
```

As you can see, we can not omit the `unknown` which previously allowed us to cast the value of `.toPrimitive()` to a `bigint` type that would never really be a possible return value, thus lying about the types.

crates/precompile-evm-to-native-swap/src/tests.rs
dmitrylavrenov: Thanks for noting, improved with `assert_noop` - [f785ce2](https://github.com/humanode-network/humanode/pull/1449/commits/f785ce2f818696c0c6a6f7a112be65ee92943849)

utils/e2e-tests/ts/tests/swap/nativeToEvm.ts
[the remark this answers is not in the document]
dmitrylavrenov: Thanks for such explanation. Now it's clear and everything sounds reasonable.

Btw, it's my stupid bad as originally it was a different comparison, then I introduced `BigInt(...)` and forgot to use an explicit conversion. Now just `toPrimitive as number` here works - https://github.com/humanode-network/humanode/pull/1449/commits/6f8ad9220b7fbb4c15fedff335f6f1a0ec798d9d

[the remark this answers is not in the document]
MOZGIII: But you don't know if it's actually a number - it could be a `string`; you can check by logging the `typeof ...` at runtime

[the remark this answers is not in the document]
dmitrylavrenov: I see, added - https://github.com/humanode-network/humanode/pull/1449/commits/bee8b5c4c641d78f999265ede5567a2c466042b3

PR #1453, opened 2025-02-21
Use gas value for simple transfer call from EVM config

dmitrylavrenov:


MOZGIII (commented):
It is better the way it is in tests, no need for this change

dmitrylavrenov:
https://github.com/humanode-network/humanode/pull/1453#pullrequestreview-2632650788

PR #1473, opened 2025-03-04
Fix rustup after it introduced changes to the toolchain installation commands

MOZGIII:


dmitrylavrenov (approved):
LGTM!

docker/build.sh
```diff
@@ -42,6 +42,7 @@ docker run \
   --rm \
   -i \
   --platform "$PLATFORM" \
+  --init \
```
dmitrylavrenov, line 45: Just one question, why should we add `init`?

I didn't come up with it when played with possible fix

MOZGIII: Since we'd now run a shell as as pid 1 instead of the builder, we need something to reliably pass signals to the child processes - and this is where `--init` helps. It will run as pid 1 instead and make sure signals (like SIGTERM) are delivered to the children properly.

PR #1498, opened 2025-03-24
Fix CLI help for `--rpc-url` parameter

quasiyoke:
CLI help for `--rpc-url` parameter doesn't match the current behavior if not passed. If you don't specify `--rpc-url`, `--rpc-url-ngrok-detect`, `--rpc-url-unset` and `--rpc-url-scheme-preference` parameters, [WebSocket will be chosen](https://github.com/humanode-network/humanode/blob/0e7f0369fee1d8f6f6c6f5f1bf93677910ff7e99/crates/humanode-peer/src/cli/config.rs#L147-L150) as the RPC URL scheme (CLI help points out HTTP). This is correctly specified in [our documentation](https://gitbook.humanode.io/docs/node/cli) (in general). The behavior was introduced in https://github.com/humanode-network/humanode/pull/355 Recent changes to 0.9.43 [haven't affected this behavior](https://github.com/humanode-network/humanode/pull/940/files#diff-e5de2ad4f6461b97bcdb868153cacd84150b70592f4332612d2f4e7bab78cb7bL156-L159) significantly.

PR #1499, opened 2025-03-24
Remove already applied `pallet_multisig::migrations::v1::MigrateToV1` at `humanode-runtime` migrations scope

dmitrylavrenov:
`try-runtime` checks at ci throw error now `called Result::unwrap() on an Err value: Other("this migration can be deleted")` after executed runtime upgrade at mainnet to 127 due to https://github.com/humanode-network/substrate/blob/8afcafd721d31939e4dee7b779f152ab9eb1c8be/frame/multisig/src/migrations.rs#L49

#1498 has faced this error. So, probably, we should remove it now. 

Btw, I don't think that `pallet_multisig::migrations` should throw error for this logic at `pre_upgrade` implementation.

### DIFF
--- a/crates/humanode-runtime/src/lib.rs
+++ b/crates/humanode-runtime/src/lib.rs
@@ -889,7 +889,6 @@ pub type Executive = frame_executive::Executive<
             ConstU32<1000>,
             frontier_precompiles::FrontierPrecompilesAddresses<Runtime>,
         >,
-        pallet_multisig::migrations::v1::MigrateToV1<Runtime>,
         storage_version_initializer::StorageVersionInitializer<Balances, Runtime>,
     ),
 >;


quasiyoke (approved):

dmitrylavrenov:
> This is actually interesting, and finally forces some pattern on the migrations management: remove them as soon as they're no longer useful. This leads to some a set of complications and new things we have to keep an account for. Let's create an issue to discuss and think this through.

Yep, that's why the following reworks was introduced - https://github.com/humanode-network/humanode/pull/1315 to unify the whole migrations scope.

dmitrylavrenov:
> This is actually interesting, and finally forces some pattern on the migrations management: remove them as soon as they're no longer useful. This leads to some a set of complications and new things we have to keep an account for. Let's create an issue to discuss and think this through.

https://github.com/humanode-network/humanode/issues/1500

PR #1507, opened 2025-03-31
Fungible conformance Inspect and Mutate test for `pallet-evm-balances`

### LINKED ISSUE #1438: Implement fungible conformance tests: Inspect and Mutate
Details: https://github.com/paritytech/substrate/pull/13852

Should be implemented after #940

dmitrylavrenov:
Closes #1438 

- [x] add the corresponding macro that runs https://github.com/humanode-network/substrate/tree/h-polkadot-v0.9.43/frame/support/src/traits/tokens/fungible/conformance_tests
- [x] adapt mock env to satisfy `AccountId: AtLeast8BitUnsigned`

### DIFF
--- a/crates/pallet-evm-balances/src/mock.rs
+++ b/crates/pallet-evm-balances/src/mock.rs
@@ -19,24 +19,20 @@ use crate::{self as pallet_evm_balances, *};
 pub(crate) const INIT_BALANCE: u64 = 10_000_000_000_000_000;
 
 /// Alice account.
-pub fn alice() -> u64 {
-    5234
-}
+pub const ALICE: u64 = 0x5234;
 
 /// Alice H160 account.
-pub fn alice_h160() -> H160 {
-    H160::from_low_u64_be(alice())
-}
+pub const ALICE_H160: H160 = H160(hex_literal::hex!(
+    "0000000000000000000000000000000000005234"
+));
 
 /// Bob account.
-pub fn bob() -> u64 {
-    4325
-}
+pub const BOB: u64 = 0x4235;
 
 /// Bob H160 account.
-pub fn bob_h160() -> H160 {
-    H160::from_low_u64_be(bob())
-}
+pub const BOB_H160: H160 = H160(hex_literal::hex!(
+    "0000000000000000000000000000000000004235"
+));
 
 /// H160 into u64 address mapper.
 pub struct H160IntoU64;
@@ -188,8 +184,8 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
                     nonce: Default::default(),
                     storage: Default::default(),
                 };
-                map.insert(alice_h160(), init_genesis_account.clone());
-                map.insert(bob_h160(), init_genesis_account);
+                map.insert(ALICE_H160, init_genesis_account.clone());
+                map.insert(BOB_H160, init_genesis_account);
                 map
             },
         },
--- a/crates/pallet-evm-balances/src/mock.rs
+++ b/crates/pallet-evm-balances/src/mock.rs
@@ -5,7 +5,7 @@ use frame_support::{
     traits::{ConstU32, ConstU64, FindAuthor},
     weights::Weight,
 };
-use pallet_evm::{EnsureAddressNever, FixedGasWeightMapping, IdentityAddressMapping};
+use pallet_evm::{AddressMapping, EnsureAddressNever, FixedGasWeightMapping};
 use sp_core::{H160, H256, U256};
 use sp_runtime::{
     generic,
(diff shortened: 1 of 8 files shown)


crates/pallet-evm-balances/src/mock.rs
```diff
@@ -18,12 +18,33 @@ use crate::{self as pallet_evm_balances, *};
 
 pub(crate) const INIT_BALANCE: u64 = 10_000_000_000_000_000;
 
-pub(crate) fn alice() -> H160 {
-    H160::from_str("1000000000000000000000000000000000000000").unwrap()
+/// Alice account.
+pub fn alice() -> u64 {
+    5234
 }
```
quasiyoke, line 24: I was surprised when I saw such functions. Then I figured out that they're used in tests in several modules. In general, I am upset that addrs are of `Copy` types, and we don't rely on it. `H160` doesn't have very convenient constructors and, AFAIK, there is no suitable macro to build it. There's a `const fn repeat_byte` constructor — but then our test addresses should fit in `u8`, not `u64`. However, they made an inner array as `pub`. I've considered this variant:

```suggestion
/// Alice account.
pub const ALICE: u64 = 0x1472; // 5234

/// Alice H160 account.
pub const ALICE_ADDR: H160 = H160([
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x14, 0x72,
]);
```

— I think that this option is still worse in terms of readability than the original.

I still would like to replace the original address numbers (`fn alice`) with `const` (without `fn alice_h160` refactoring) 189137d49942956877d8286f8761acf7bd85a91c

quasiyoke (approved):
I think the `run_tests!` macro suits well, maybe it would be worth adding something like that to the SDK.

crates/pallet-evm-balances/src/tests/mod.rs
```diff
@@ -42,8 +42,9 @@ fn basic_setup_works() {
 fn evm_system_removing_account_non_zero_balance() {
     new_test_ext().execute_with_ext(|_| {
         // Prepare test preconditions.
-        let contract = H160::from_str("1000000000000000000000000000000000000003").unwrap();
-        EVM::create_account(contract, vec![1, 2, 3]);
+        let contract = 3;
+        let contract_h160 = H160::from_low_u64_be(contract);
```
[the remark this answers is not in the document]
dmitrylavrenov: The contract variable is used at transfer call. So, we can't just remove it.

crates/pallet-evm-balances/src/mock.rs
dmitrylavrenov: Thanks, I used an old code here. Constants are much better. Will use it.

dmitrylavrenov: Thanks for noting, let me think a little bit about it.

MOZGIII: There is hex_literal crate/macro that we'd usually use

MOZGIII: As I understand, we need to use `u64` instead of H160 here brcause it is what the test macros assume.

crates/pallet-evm-balances/src/tests/fungible_conformance_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+use paste::paste;
+
+use crate::{mock::*, *};
+
+macro_rules! run_tests {
+    ($path:path, $($name:ident),*) => {
+		$(
+			paste! {
+				#[test]
+				fn [< $name _dust_trap_on >]() {
```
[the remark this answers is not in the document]
MOZGIII: Not sure, but maybe this will be of use for what you have in mind: https://github.com/danielhenrymantilla/macro_rules_attribute-rs

crates/pallet-evm-balances/src/mock.rs
dmitrylavrenov: > As I understand, we need to use u64 instead of H160 here brcause it is what the test macros assume.

Right

crates/pallet-evm-balances/src/tests/fungible_conformance_tests.rs
```diff
@@ -0,0 +1,66 @@
+use frame_support::traits::fungible::{conformance_tests, Inspect, Mutate};
+use paste::paste;
+
+use crate::{mock::*, *};
+
+macro_rules! run_tests {
```
MOZGIII, line 6: Can't we import this macro from the `conformance_tests::inspect_mutate`?

MOZGIII: We can't:
- it's not there but in a different place
- it's not expored
- it's not generic enough to be reusable like this

MOZGIII (approved):

crates/pallet-evm-balances/src/mock.rs
dmitrylavrenov: Improved with constants usage

crates/pallet-evm-balances/src/tests/fungible_conformance_tests.rs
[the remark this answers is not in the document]
quasiyoke: Interesting crate! Although it seems to be rarely contributed to. I tried it in this context, it would be more convenient to use if there was a partial application, like:

```rs
#[test]
#[apply(rename_fn, $name, _dust_trap_on)]
fn dust_trap_on {
```

By the way, I found the [`std_tools::rename_fn` macro](https://docs.rs/impls_index/0.1.3/src/impls_index/func.rs.html#38), but the macro is not the right one, and I don't really like the crate.

PR #1561, opened 2025-06-20
CI: "upload runtime" step

quasiyoke:


.github/actions/plan/modes.ts
```diff
@@ -63,10 +64,28 @@ export const code = {
...
+  //   cargoArgs: "--locked --workspace --release --package humanode-runtime",
+  //   cargoCacheKey: "runtime",
+  //   platformIndependent: true,
+  //   artifactSelector: "runtime",
+  // },
+  buildRuntimeEvmTracing: {
+    name: "build runtime with EVM tracing",
+    cargoCommand: "build",
+    cargoArgs:
+      "--locked --workspace --release --package humanode-runtime --features evm-tracing",
```
MOZGIII, line 86: Do not use `--package` if possible as it can scramble features is a way we not checking for (we are not ready to use this kind of target selection until we add CI tests for this), use the `--bin` (or similar) target instead

.github/scripts/artifact-upload-params.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+RUNTIME_PATH='target/release/wbuild/humanode-runtime/humanode_runtime.compact.compressed.wasm'
+
+case "$ARTIFACT_SELECTOR" in
```
MOZGIII, line 6: Selector should only be responsible for selecting the artifact path, not for naming the artifact (that can be addressed separately - like by adding another knob to the mode)

MOZGIII:
> I believe we can remove `cfg...feature` here to include `substrate_wasm_builder` by default in no-std:
> 
> https://github.com/humanode-network/humanode/blob/adab1b49783d96c9195a748d4c6f7283098aad26/crates/humanode-runtime/build.rs#L8
> 
> — and then we will be able to build WASM using the following command:
> 
> ```
> cargo build --locked --workspace --release --package humanode-runtime --no-default-features --features evm-tracing --target wasm32-unknown-unknown
> ```
> 
> This would reduce the number of dependency crates to build from the current 1270 to just 449. The only thing is, I haven't tested it yet to see if it'll break anything else.

If we remove it it will trigger a wasm build while doing a wasm build - an infinite loop

MOZGIII:
After some research, I deem the way of building the runtime via the `WasmBuilder` preferable; at present that would mean building via `build.rs` script. It we could avoid building the dependencies of the `humanode-runtime` before the build script it executed is the usual `std` mode - then we will have an advantage of not having to pay the price for building them twice...

dmitrylavrenov (approved):
Clear on my side.

Not sure, that I'm in context having a lot of commented code. Is it fully ready for merging or just some kind of checkpoint ?

.github/actions/plan/modes.ts
quasiyoke: I think it will be ideal for us to build with the `--lib` flag; I almost forgot it exists, I’ll add it.

quasiyoke:
@MOZGIII 

> I deem the way of building the runtime via the `WasmBuilder` preferable; at present that would mean building via `build.rs` script.

I noticed that Polkadot itself sometime uses wasm-builder from [helper scripts](https://github.com/paritytech/polkadot-sdk/blob/cc5065dab6ab4c40d2901981481afdad8e34d4c2/substrate/.maintain/build-only-wasm.sh). I also believe it's easier to build from `build.rs`, especially since that was the method envisioned in [their template](https://github.com/paritytech/polkadot-sdk-minimal-template/blob/master/runtime/Cargo.toml), and we don’t need any specific customization yet. 

> If we could avoid building the dependencies of the `humanode-runtime` before the build script it executed for the usual `std` mode - then we will have an advantage of not having to pay the price for building them twice (build script will only build the dependencies once, for wasm, and here are the savings)...

I don’t yet know how to do this.

quasiyoke:
@dmitrylavrenov 

> [what about] having a lot of commented code. Is it fully ready for merging or just some kind of checkpoint?

We agreed with Mozgiii right away that launching CI here would require several attempts, and that running too many jobs would consume a significant portion of our CI budget. Therefore, in the first commit, I disabled all CI jobs except those necessary for demonstrating the new artifact type. I think we should first ensure that there are no comments on the Draft PR, and then I'll remove this commit in the final PR.

.github/actions/plan/modes.ts
```diff
@@ -4,6 +4,8 @@ export type Mode = {
   cargoArgs: string;
   cargoCacheKey: string;
   platformIndependent?: true;
+  artifactSelector?: "peer" | "runtime";
+  artifactFeaturesMarker?: "evm-tracing";
```
MOZGIII, line 8:
```suggestion
  artifactMarker?: string;
```

More generic - doesn't have to be specifically about the features, and can easily be any string

.github/workflows/build.yml
```diff
@@ -98,6 +98,7 @@ jobs:
         env:
           PLATFORM_ARTIFACT_MARKER: ${{ matrix.plan.platform.artifactMarker }}
           ARTIFACT_SELECTOR: ${{ matrix.plan.mode.artifactSelector }}
+          FEATURES_MARKER: ${{ matrix.plan.mode.artifactFeaturesMarker }}
```
MOZGIII, line 101:
```suggestion
          ARTIFACT_MARKER: ${{ matrix.plan.mode.artifactMarker }}
```
Better use the same naming everywhere

quasiyoke: Certainly, that will be more systematic

.github/scripts/artifact-upload-params.sh
```diff
@@ -0,0 +1,40 @@
...
+set -euo pipefail
+
+case "$ARTIFACT_SELECTOR" in
+  peer)
+    ARTIFACT_PATH='target/release/humanode-peer'
+    ARTIFACT_NAME="humanode-peer-$(rustc -vV | sed -n 's|host: ||p')"
+
+    if [[ "$PLATFORM_ARTIFACT_MARKER" != "" ]]; then
+      ARTIFACT_NAME="${ARTIFACT_NAME}-${PLATFORM_ARTIFACT_MARKER}"
+    fi
```
MOZGIII, line 11: I'm thinking this section and the `MODE_ARTIFACT_MARKER` should be applied universally for all artifact selectors

.github/actions/plan/modes.ts
MOZGIII: Don't see the type change to `string`

.github/scripts/artifact-upload-params.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+case "$ARTIFACT_SELECTOR" in
+  peer)
```
MOZGIII, line 5: I think it would make more sense for this to be the default selector mode for this file, and not leak beyond it; then we don't change the semantics for the existing CI jobs (which we arguable shouldn't do here)

.github/scripts/artifact-upload-params.sh
```diff
@@ -0,0 +1,40 @@
...
+    if [[ "${PATHEXT:-""}" != "" ]]; then
+      ARTIFACT_PATH="${ARTIFACT_PATH}.exe"
+    fi
+    ;;
+
+  runtime)
+    if [[ -z "${MODE_ARTIFACT_MARKER}" ]]; then
+      printf 'MODE_ARTIFACT_MARKER must not be empty\n' >&2
+      exit 1
+    fi
```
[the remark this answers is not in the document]
quasiyoke: ✅

.github/scripts/artifact-upload-params.sh
MOZGIII: We could make it changeable in the mode - but let's leave it for now, it is good enough

.github/scripts/artifact-upload-params.sh
MOZGIII, line 1: Would be nice to have the syntax and formatting unified; I recommend using `Bash IDE` and `shellcheck` extensions

quasiyoke: `shellcheck` — I had it installed,
`BashIDE` — I installed it, but it didn't offer any formatting corrections…

.github/actions/plan/modes.ts
```diff
@@ -4,6 +4,8 @@ export type Mode = {
   cargoArgs: string;
   cargoCacheKey: string;
   platformIndependent?: true;
+  artifactSelector?: "runtime";
+  artifactMarker?: "evm-tracing";
```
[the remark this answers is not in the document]
quasiyoke: ✅

quasiyoke:
I removed the CLI arguments `--package humanode-runtime --lib` – the number of crates during compilation remained at 1270. The included std feature likely prevented limiting the scope of compiled packages. It turns out `--package` only posed a potential risk of breaking features without providing any performance benefits.

I will still see what can be done to avoid building so many crates.

MOZGIII:
Maybe we could just structure the builds in such a way that we also publish a whole node with tracing support? I.e. so that the node would, like, also have the embedded tracing runtime initially, and the tracing variant of the on-chain runtime (code).

quasiyoke:
An experiment confirmed that the cargo flags `--package` and `--workspace` work exactly as @Mozgiii initially explained:
`--package` builds the selected package with the minimal possible set of features.
`--bin` builds the selected binary with a set of features for a complete workspace build.
`--workspace` builds the entire workspace in any case.
— these options work this way both in 2021 and 2024 edition, with resolver `2` and `3` respectively.

Added Cargo target `--bin humanode-runtime-build-selector` — the number of crates necessary for building is 673, compared to 1270 when using `--workspace`.

> also publish a whole node with tracing support

I think a separate WASM artifact is more flexible to use. If it’s not needed, I can add node with EVM-tracing here. If another artifact (EVM-tracing node) is required, perhaps it would be best to create it in a separate task?

crates/humanode-runtime/src/bin/build_selector.rs
MOZGIII, line 1: I think it like this will still build all the dependencies for the host triple before the actual runtime is built (for the wasm triple).

If we, instead, make another crate, and have the contents of the `build.rs` script in there (with slight changes to target a different crate) - we could essentially directly trigger the runtime build for wasm from a separate package - as opposed to doing it from the build script.

MOZGIII: > If we, instead, make another crate, and have the contents of the `build.rs` script in there (with slight changes to target a different crate) - we could essentially directly trigger the runtime build for wasm from a separate package - as opposed to doing it from the build script.

Tested this with https://github.com/humanode-network/humanode/commit/842c319b259dd28a541b055cc8a3bcfb94239c23 - the build script fails (`substrate-wasm-builder` panics at https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/utils/wasm-builder/src/wasm_project.rs#L837); the other variation where the build script it the bin itself also didn't work - but for some other reason.

PR #1565, opened 2025-06-24
[evm-tracing-2] сlient-side implementation

dmitrylavrenov:


crates/evm-tracing-client/src/types/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+/// Memory converter.
+pub fn convert_memory(memory: Vec<u8>) -> Vec<H256> {
+    let size = 32;
+    memory
+        .chunks(size)
+        .map(|c| {
+            let mut msg = [0u8; 32];
+            let chunk = c.len();
+            if chunk < size {
+                let left = size.checked_sub(chunk).expect("valid op; qed.");
```
MOZGIII, line 105:
```suggestion
            if let Some(left) = size.checked_sub(chunk) {
```

crates/evm-tracing-client/src/serialization.rs
```diff
@@ -0,0 +1,102 @@
...
+{
+    if let Some(value) = value.as_ref() {
+        return string_serialize(value, serializer);
+    }
+
+    Err(S::Error::custom("string serialize error."))
+}
+
+/// Serializes `U256`.
+pub fn u256_serialize<S>(data: &U256, serializer: S) -> Result<S::Ok, S::Error>
```
quasiyoke, line 89: Possibly, it's worth noting in the name that serialization could happen with saturation, e.g.: `u256_as_saturated_u64` Checked out all the places where this is applied, and saturation shouldn't occur, but someone might use it in the future where saturation *could* happen.

crates/evm-tracing-client/src/serialization.rs
```diff
@@ -0,0 +1,102 @@
...
+where
+    S: Serializer,
+{
+    serializer.serialize_u64(UniqueSaturatedInto::<u64>::unique_saturated_into(*data))
```
quasiyoke, line 93: Perhaps it's worth considering converting with a `try_from` and explicitly returning an error in case of conversion issues?

crates/evm-tracing-client/src/serialization.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/// Serializes `H256` 0x.
+pub fn h256_0x_serialize<S>(data: &H256, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    serializer.serialize_str(&format!("0x{:x}", data))
```
quasiyoke, line 101: This might require less allocations if the serializer has a custom implementation:

```suggestion
    serializer.collect_str(&format_args!("0x{data:x}"))
```

crates/evm-tracing-client/src/types/call_tracer.rs
```diff
@@ -0,0 +1,90 @@
...
+        call_type: Vec<u8>,
+        /// Input.
+        #[serde(serialize_with = "bytes_0x_serialize")]
+        input: Vec<u8>,
+        /// To.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        to: Option<H160>,
+        #[serde(
+            skip_serializing_if = "Option::is_none",
+            serialize_with = "option_bytes_0x_serialize"
```
quasiyoke, line 67: Wouldn't we like to use `serde_with`? It would eliminate the need for "ad hoc" serializers.

```rs
#!/usr/bin/env rust-script
//! ```cargo
//! [dependencies]
//! hex = "0.4"
//! serde = { version = "1", features = ["derive"] }
//! serde_json = "1"
//! serde_with = "3.14"
//! ```

use std::marker::PhantomData;

use serde::{Serialize, Serializer};
use serde_with::{SerializeAs, serde_as};

#[serde_as]
#[derive(Serialize)]
struct Create {
    #[serde_as(as = "PrefixedHex<Vec<u8>>")]
    input: Vec<u8>,
    #[serde_as(as = "Option<PrefixedHex<Vec<u8>>>")]
    output: Option<Vec<u8>>,
}

struct PrefixedHex<T> {
    _marker: PhantomData<T>,
}

impl SerializeAs<Vec<u8>> for PrefixedHex<Vec<u8>> {
    fn serialize_as<S: Serializer>(bytes: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error> {
        serializer.collect_str(&format_args!("0x{}", hex::encode(bytes)))
    }
}

fn main() {
    let r = Create {
        input: vec![0xca, 0xfe],
        output: Some(vec![0xba, 0xbe]),
    };
    println!("{}", serde_json::to_string(&r).unwrap());
}
```

A `PrefixedHex` could have a similar impl for `U256`.

dmitrylavrenov: Would like to keep it as is to simplify applying the diffs later without introducing new crate.

crates/evm-tracing-client/src/types/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+pub enum ContextType {
+    /// Call type.
+    Call(CallType),
+    /// Create type.
+    Create,
+}
+
+impl ContextType {
+    /// Obtain context type from opcode.
+    pub fn from(opcode: Vec<u8>) -> Option<Self> {
```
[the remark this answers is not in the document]
dmitrylavrenov: Looks like it can be addressed at https://github.com/humanode-network/humanode/issues/1600

crates/evm-tracing-client/src/serialization.rs
dmitrylavrenov: I prefer to have the naming in the same way it's used in moonbeam now to simplify applying the diffs later

crates/evm-tracing-client/src/formatters/trace_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+                                    output,
+                                })
+                            }
+                            CallResult::Error(error) => TransactionTraceOutput::Error(error),
+                        },
+                        subtraces: trace.subtraces,
+                        trace_address: trace.trace_address.clone(),
+                        // Can't be known here, must be inserted upstream.
+                        transaction_hash: H256::default(),
+                        // // u32 (eth tx index) is big enough for this truncation to be practically impossible.
```
quasiyoke, line 67: (minor) commented twice

crates/evm-tracing-client/src/listeners/call_list.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// See `RuntimeEvent::StepResult` event explanatioins.
+    step_result_entry: Option<(u32, Call)>,
+    /// When tracing a block `Event::CallListNew` is emitted before each Ethereum transaction is
+    /// processed. Since we use that event to **finish** the transaction, we must ignore the first
+    /// one.
+    call_list_first_transaction: bool,
+    /// True if only the `GasometerEvent::RecordTransaction` event has been received.
+    /// Allow to correctly handle transactions that cannot pay for the tx data in Legacy mode.
+    record_transaction_event_only: bool,
+}
```
quasiyoke, line 63: I didn't understand why we don't need [`with_log`](https://github.com/moonbeam-foundation/moonbeam/blob/4a7944905aa8b6d5e897e2c1f209f4468747b379/client/evm-tracing/src/listeners/call_list.rs#L84). Reviewed the discussion history. Did we decide not to support `EvmEvent::Log`?

quasiyoke (approved):

crates/evm-tracing-client/src/listeners/call_list.rs
dmitrylavrenov: Please, check the `evm` version we use. It doesn't contain `Log` logic. 

Additionally, we can't just bump it as it should be compatible with our frontier fork.

dmitrylavrenov: https://github.com/rust-ethereum/evm/blob/b7b82c7e1fc57b7449d6dfa6826600de37cc1e65/src/tracing.rs#L14

MOZGIII (approved):

PR #1566, opened 2025-06-26
CI build: improve word order in artifact marker naming

quasiyoke:
Related: #1561


MOZGIII:
This is not the "marker of the platform" it is, rather, a "marker specified in the platform definition". Can be anything since platforms themselves are not necessarily semantically linked to artifacts. Hence the term marker in the first place.

quasiyoke:
Makes sense

PR #1574, opened 2025-07-19
[evm-tracing-0] add a corresponding e2e tests

dmitrylavrenov:


utils/e2e-tests/bash/run-tests
```diff
@@ -14,7 +14,10 @@ export RUNTIME_WASM_PATH
...
 for TESTFILE in "$SCRIPT_DIR"/tests/"$TEST_SUITE"/*.sh; do
-  printf "=> Running test %s\n" "$TESTFILE" >&2
-  "$TESTFILE" "$@"
-  printf "=> Test %s passed\n" "$TESTFILE" >&2
+  # There are no tests for some test suits.
+  if [[ -f "$TESTFILE" ]]; then
+    printf "=> Running test %s\n" "$TESTFILE" >&2
+    "$TESTFILE" "$@"
+    printf "=> Test %s passed\n" "$TESTFILE" >&2
+  fi
```
MOZGIII, line 22: Shouldn't be possible... Am I missing somerthing?

MOZGIII: Have you tested it? Maybe I'm wrong here...
But what exactly is this test condition supposed to check? What the `TESTFILE` value will be?

utils/e2e-tests/ts/run-tests
```diff
@@ -10,6 +10,6 @@ shift
 HUMANODE_PEER_PATH="$(realpath -s "$HUMANODE_PEER_PATH")"
 export HUMANODE_PEER_PATH
 
-if [[ "$TEST_SUITE" == "base" ]]; then
-  exec yarn utils/e2e-tests/ts test
+if [[ -d "$SCRIPT_DIR/tests/$TEST_SUITE"/ ]]; then
```
MOZGIII, line 13: Not sure we need this check; doesn't it make sense to fail if a non-existent test suite is requested?

MOZGIII: For the maintenance purposes - I'd say this definitely makes sense... Once can easily miss some place to be updated upon renames - we'd want this to be very visible

utils/e2e-tests/ts/tests/evm-tracing/debugTraceCall.ts
```diff
@@ -0,0 +1,3 @@
+import { test } from "vitest";
+
+test("debug trace call", () => {});
```
MOZGIII, line 3:
```suggestion
test("todo", () => {});
```

utils/e2e-tests/bash/run-tests
dmitrylavrenov: Sure, I tested it. 

First time, when I added `evm-tracing` test suite, I met an error that there are no files under `evm-tracing` to run the test. 

That's why I added this check to avoid having the error.

utils/e2e-tests/ts/run-tests
MOZGIII: Is this bad? Does it actually block anything? What if allow tests to fail if there are no tests to run?

MOZGIII: I suppose we want to avoid spawning CI jobs when they do no useful work...

MOZGIII (approved):

utils/e2e-tests/ts/run-tests
dmitrylavrenov: > Is this bad? Does it actually block anything? What if allow tests to fail if there are no tests to run?

The whole CI job is failed as 1 code error is returned

MOZGIII: Makes sense actually - why would we want a I job **not** to fail if no tests were run? We will just *not run the job* instead :D

MOZGIII: What do you think? Let's make the test suites / subsystems follow this logic?

MOZGIII: Yes, kind of!

But better yet, express it more explicitly via the matrix itself:

```yaml
          - suite: "base"
            build-args: ""
          - suite: "try-runtime"
            build-args: "--features try-runtime"
            skip-ts-tests: true
          - suite: "evm-tracing"
            build-args: "--features evm-tracing"
            skip-bash-tests: true
```

```yaml
if: matrix.test.skip-ts-tests != true
```

utils/e2e-tests/bash/run-tests
dmitrylavrenov: Not relevant now

utils/e2e-tests/ts/run-tests
dmitrylavrenov: Improved

Thanks for the idea, now it's much better!

dmitrylavrenov:
Will rebase after #1580

PR #1580, opened 2025-07-24
Introduce `skip-bash-tests` and `skip-ts-tests` flags to control e2e-tests jobs execution at CI

dmitrylavrenov:
As a follow up of work on https://github.com/humanode-network/humanode/pull/1574

### DIFF
--- a/utils/e2e-tests/ts/run-tests
+++ b/utils/e2e-tests/ts/run-tests
@@ -10,4 +10,4 @@ shift
 HUMANODE_PEER_PATH="$(realpath -s "$HUMANODE_PEER_PATH")"
 export HUMANODE_PEER_PATH
 
-exec yarn utils/e2e-tests/ts test "./${TEST_SUITE}/"
+exec yarn utils/e2e-tests/ts test "./tests/${TEST_SUITE}/"
--- a/utils/e2e-tests/ts/run-tests
+++ b/utils/e2e-tests/ts/run-tests
@@ -10,4 +10,4 @@ shift
 HUMANODE_PEER_PATH="$(realpath -s "$HUMANODE_PEER_PATH")"
 export HUMANODE_PEER_PATH
 
-exec yarn utils/e2e-tests/ts test "$TEST_SUITE"
+exec yarn utils/e2e-tests/ts test /"$TEST_SUITE"/
--- a/utils/e2e-tests/ts/run-tests
+++ b/utils/e2e-tests/ts/run-tests
@@ -10,4 +10,4 @@ shift
 HUMANODE_PEER_PATH="$(realpath -s "$HUMANODE_PEER_PATH")"
 export HUMANODE_PEER_PATH
 
-exec yarn utils/e2e-tests/ts test /"$TEST_SUITE"/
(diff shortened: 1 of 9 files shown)


utils/e2e-tests/ts/run-tests
```diff
@@ -10,6 +10,4 @@ shift
 HUMANODE_PEER_PATH="$(realpath -s "$HUMANODE_PEER_PATH")"
 export HUMANODE_PEER_PATH
 
-if [[ "$TEST_SUITE" == "base" ]]; then
-  exec yarn utils/e2e-tests/ts test
-fi
+exec yarn utils/e2e-tests/ts test /"$TEST_SUITE"/
```
dmitrylavrenov, line 13: Added slashes to apply tests filter by having a corresponding `test suite` directory in the test file path.

MOZGIII: This is rather odd... I mean, the first `/`. Do we tell the test command to do a lookup from the root here?

MOZGIII: I'm just trying to interpret the meaning and justification here. What is the issues if we don't have the slashes?

dmitrylavrenov: Maybe, the first `/` is not required.

Let me explain the goal of this. I added the following structure for testing
```
- tests
  - base
    - hello.ts
    - world.ts
  - hello
    - base.ts
```

In case simple run of  `exec yarn utils/e2e-tests/ts test base`, the following tests are run: `base.ts`, `hello.ts`, `world.ts`.

So, we should add slash. Tested with both cases: `/"$TEST_SUITE"/` `$TEST_SUITE"/` - works.

dmitrylavrenov: Just tested with my sample, it doesn't work. Tried `hello` test suit. it runs both `hello/base.ts` and `base/hello.ts`

dmitrylavrenov: https://github.com/humanode-network/humanode/pull/1580/commits/a83e0b37a9dc5324e8162ef46a5cf7c0aabb357f

dmitrylavrenov: @MOZGIII What do you think about current version ? is it ok ?

MOZGIII: > So, we essentially want to pass `/base/` there? Not `/base`, not `./base`, not `base/`, and not `./base/`?
> 
> What if we have `a/b/base/mytest.ts`?

Does it work for all the cases, even with a deeply nested dir with the same name?

MOZGIII: We can try using https://vitest.dev/config/#dir option

quasiyoke (approved):

utils/e2e-tests/ts/run-tests
dmitrylavrenov: Hm.. Added additional example like `tests/base/tests/hello/hello.ts` containing deeply `tests`.  It's taken as well in case `hello` test suite.

